PR #41, opened 2014-09-29
Update install-windows.md

wetdog:
Yesterday I find the problem with websockets when installing tidal.


kindohm:
Propose we close this. @yaxu @wetdog? I've installed on Windows many times successfully since this issue was opened and have not been able to reproduce the problem.

PR #46, opened 2014-11-04
Update install-linux.sh

louisduhamel:
Just to make sure that cabal is up to date (it was not the case when I used the script last week).


yaxu:
Merged for now, sorry this took so long..

PR #48, opened 2015-02-22
two typo parenthesis

radames:


lennart:
Thanks @radames, but we moved the latest tidal documentation into a [separate repository](https://github.com/tidalcycles/tidalcycles.github.io).

@alex didn't we want to remove tidal.md from this repo? Apart from that, this pull request shows what is missing from the current documentation. The content is already [in the sources](https://github.com/tidalcycles/tidalcycles.github.io/blob/master/_patterns/mapping-over.md) but their categories are wrong. I am unsure if we should put some of the basics in to the "Getting Started"-Chapter or make them subsections of [patterns/sequences](http://tidal.lurk.org/patterns.html#sequences).

ideas?

radames:
hi @lennart, perfect! I'll check it out!

yaxu:
Hey @lennart I think mapping is best as a subset of pattern transformers in the reference.

We should reorganise things at some point though. A few things could be moved around, e.g. synth parameters shouldn't be under continuous patterns, and jux shouldn't really be under compositions.

lennart:
I'd suggest using this issue tracker for managing the reorganization. Maybe we should add a new topic on lurk to bundle general discussion about the documentation (or just link to the issues from the original -> rescue for a complete documentation)

I just moved [Mapping over patterns](http://tidal.lurk.org/functions.html#mapping-over-patterns <http://tidal.lurk.org/functions.html#mapping-over-patterns>) in to the pattern transformers section. I’ll open tickets here for everything else…

PR #52, opened 2015-04-01
preplace/protate

unthingable:
Pattern convolver/quantizer. Not sure this is the best name, happy to wait for a recommendation for a better one.

http://lurk.org/groups/tidal/messages/topic/5XKHejBaGvjdw0dksJ6xGs

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -543,7 +543,7 @@ preplace = prr 0 (1, 1)
 
 -- | @protate len rot p@ rotates pattern @p@ by @rot@ beats to the left.
 -- @len@: length of the pattern, in cycles.
--- Example: @d1 $ every 4 (protate 2 -1) $ slow 2 $ sound "bd hh hh hh"@
+-- Example: @d1 $ every 4 (protate 2 (-1)) $ slow 2 $ sound "bd hh hh hh"@
 protate :: Time -> Int -> Pattern a -> Pattern a
 protate len rot p = prr rot (len, len) p p
 
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -511,11 +511,9 @@ prrw f rot (blen, vlen) beatPattern valuePattern =
     (arc (density cycles $ beatPattern) (0, blen))
     (drop (rot `mod` length values) $ cycle values)
 
-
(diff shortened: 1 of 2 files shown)


unthingable:
Added event shifting with `protate` and `<<~`/`~>>` companion operators. This is starting to need tests. :)

bgold-cosmos:
Just to let you know, I am looking at this - I had a question about the code comments but it looks like your recent update addressed it :)  Hope to play with this a lot soon.

unthingable:
@bgold-cosmos cool! :)

I think I'm done with the additions for now, unless someone points something out.

yaxu:
I'm just wondering if there is a name that fits better than 'protate'. Something like 'structureWith', but shorter

unthingable:
Yes, I've wondered that too, both the "p" and "replace" parts. By the way, there is now a "*With" version that accepts a value combinator. :)

The "p" stands for pattern — this came during a transition from a list based function to a pattern based one. "Replace" is what's actually happening on the low level, but that's not a great description of the effect.

What about `pzip`? The effect is a kind of a doubly-asymmetric zip (discarding values from the left and timing from the right), and it has a very pattern-specific behavior. Then `preplaceWith` would become `pzipWith`, which would be semantically consistent with `zip`/`zipWith` (and be more similar to `zipWith` than `pzip` is to `zip`).

yaxu:
Yes pzip sounds better but lets think a bit longer if you don't mind, I'm on holiday this week and not doing much thinking!
Maybe `rotate` is better than `protate`? Or if the collision with Data.Bits.rotate is a potential problem, just `rot`? Or maybe just define the operators, as with `<~` and `~>`..

unthingable:
Don't mind at all. Not very concerned with Data.Bits.rotate collision either. 

`rot` sounds pretty good. The `~>>` and `<<~` operators assume unit cycles, so it's still good to have a function (one could write `rot l n = slow l . (n <<~) . density l`)

yaxu:
Ok I'm rather tired, and got a bit lost with git, and ended up merging it with the master branch, under slightly mysterious conditions. Could you please check I haven't messed this up?

unthingable:
Looks fine! Thank you.

My emacs took the initiative to trim some trailing spaces, that probably contributes to the mystery.

So in the end you decided against `pzip`/`rot`?

PR #54, opened 2015-05-06
Dirt loop

bgold-cosmos:
Adds support for a new dirt parameter "loop", which will loop the sample the specified number of times.  This respects the "begin" and "end" parameters as well, so you can loop over a portion of a sample too.  For example `d1 $ sound "sn" |+| loop "4" |+| end "0.1"` will loop over the first tenth of the sample four times.  This requires an updated Dirt.

`striateL` and `striateL'` are new functions that work just like `striate` and `striate'`, but with an extra parameter for looping.  So `striateL' 8 (1/32) 4 $ sound "rave"` chops the sample into 8 pieces, looping each four times.

### DIFF
--- a/Sound/Tidal/Dirt.hs
+++ b/Sound/Tidal/Dirt.hs
@@ -45,7 +45,7 @@ dirt = OscShape {path = "/play",
                             F "bandf" (Just 0),
                             F "bandq" (Just 0),
                             S "unit" (Just "rate"),
-                            I "loop" (Just 0)
+                            I "loop" (Just 1)
                           ],
                  cpsStamp = True,
                  timestamp = MessageStamp,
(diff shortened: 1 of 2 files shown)


Sound/Tidal/Dirt.hs
```diff
@@ -163,4 +166,12 @@ striateO :: OscPattern -> Int -> Double -> OscPattern
 striateO p n o = cat $ map (\x -> off (fromIntegral x) p) [0 .. n-1]
   where off i p = p |+| begin ((atom $ (fromIntegral i / fromIntegral n) + o) :: Pattern Double) |+| end ((atom $ (fromIntegral (i+1) / fromIntegral n) + o) :: Pattern Double)
 
+striateL :: Int -> Int -> OscPattern -> OscPattern
+--striateL n l p = cat $ map (\x -> off (fromIntegral x) p) [0 .. n-1]
```
lennart: Is there a reason why we should add the comments here?

bgold-cosmos: Nope, that's a leftover I missed from an obsolete branch.  Deleted.

PR #56, opened 2015-06-14
add stack.yaml with the latest deps

unthingable:
[stack](https://github.com/commercialhaskell/stack) is out and that's exciting!


yaxu:
Stack is new to me but sounds good..

PR #59, opened 2015-08-07
Linux installation script improvements

tpltnt:
Hi,

i improved the Linux installation script to work as an updater if run multiple times and to fail hard if problems are encountered.
Cheers,
tpltnt


yaxu:
Brilliant, thanks!

PR #60, opened 2015-08-17
added sqrt to envEq & envEqR for equal power crossfade

Djoust:
see: http://lurk.org/groups/tidal/messages/topic/iK20Gz3f6fPMJCdnLsilJ/


yaxu:
Thanks a lot!

PR #62, opened 2015-09-23
replaced all the |+| with |*|

Djoust:
is swoped all the |+| with |*|, tested and worked
the only problem is that if you use a function like stut for example in combination with |+| it gives some problems
like this:
d1 $ stut 8 0.5 0.25 $ sound "bd sn {~ bd} sn" |+| gain 0.9

maybe |+| should be removed entirely

### DIFF
--- a/Sound/Tidal/Dirt.hs
+++ b/Sound/Tidal/Dirt.hs
@@ -1,5 +1,5 @@
 {-# LANGUAGE NoMonomorphismRestriction #-}
-  
+
 module Sound.Tidal.Dirt where
 
 import Sound.OSC.FD (Datum)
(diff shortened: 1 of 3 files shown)


yaxu:
Hey Djoust, 
We should test all these, don't think it will do the right thing with any that involve pan, and probably not weave in some cases.

A bit annoying that * multiplies but + replaces here, not sure what to do about that..

kindohm:
should this PR be closed since this work is being done in the 0.6-dev branch?

yaxu:
Yes I'll close it, and will change `stut` to use `|*|` in 0.6-dev.. The others need to be thought about

PR #64, opened 2015-10-04
Make host and port of Dirt streams configurable from env variables

munshkr:
This allows Tidal to connect to a Dirt server on a different host or port with the following environment variables:
  - DIRT_HOST: Set Dirt host/ip (default: 127.0.0.1)
  - DIRT_PORT: Set Dirt port (default: 7771)

I used to override the `dirtStream` function on my bootstrap script on vim-tidal for livecoding with a Dirt server on a different host, but being able to configure it via env variables is nicier :)


yaxu:
Oops missed this, can you please resubmit this on a dev branch, thanks!

PR #69, opened 2015-11-01
update stack resolver

sbl:
the previous version was bound to ghc-7.8.4 this would also update ghc
to 7.10.2

See http://www.stackage.org/lts-3.11


yaxu:
Sorry for the delay, I didn't notice this here..

PR #79, opened 2016-01-17
strategies: jux' relative pan

telephon:
instead of overriding the inner pan value, jux’ now adds to it,
offsetting the pan position relative to it.


telephon:
already implemented in master

PR #80, opened 2016-01-17
another variant of jux

telephon:
just see which is better ...


telephon:
already implemented in master

PR #82, opened 2016-01-19
step'

bgold-cosmos:
Allows a list of samples to be used, instead of just indexing off a
single name.

Example: `step’ [“bd”, “sn:2”] “0-1-001-“


calumgunn:
This is great!

PR #97, opened 2016-03-02
preplace fix

bgold-cosmos:
Again tweaking some signatures, so the parser can figure out how to
interpret the Pattern arguments.  Functionality should be the same as
before.

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -710,7 +710,7 @@ unwrap' pp = Pattern $ \a -> arc (stack $ map scalep (arc pp a)) a
   where scalep ev = compress (fst' ev) $ thd' ev
 
 -- removes events from pattern b that don't start during an event from pattern a
-mask :: Pattern a -> Pattern b -> Pattern b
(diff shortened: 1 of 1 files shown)


lennart:
maybe we should also (if that is possible) apply the same _fix_ to ask as you have to specify the type of the first pattern there as well.

bgold-cosmos:
Sorry, can you rephrase?  I'm not sure I understand.

lennart:
yes of course.
`mask` the function used in `fit` also has the problem you fixed here for `preplace`. namely the user currently has to do:

```haskell
d1 $ sound (mask ("1 ~" :: Pattern Bool) "bd sn")
```

since mask's signature is `Pattern a -> Pattern b -> Pattern b` the first pattern's type cannot be inferred.

WHat I wondered is, whether we can also change the signature of `mask` so that the first pattern is not _that_ general anymore, e.g. Pattern String/ Pattern Bool.

bgold-cosmos:
OK, I see.  I added a commit to change it to Pattern String, since this keeps it the most general.  For some reason this seems to have created a conflict (I don't understand git sometimes...)

PR #98, opened 2016-03-03
documentation additions and merge from tidal.lurk.org

lennart:
adds docs from markdown sources currently used in tidal.lurk.org.

also adds facade modules mapping functions to reference documentation chapters. this is a cheap way to ease auto generating haddock documentation for specific functions without restructuring all of tidal core right now.

some conventions apply to successfully autogenerate contents for http://tidal.lurk.org/functions.html 

all variations of functions e.g. `jux`, `jux'` and `juxBy` have their separate haddock comment block. In this case docs will have a section `jux` with three function signatures (`jux`, `jux'`,`juxBy`) and in same order the docs for each function

code blocks _need_ the haddock example notation

    @    
    d1 $ sound "bd"
    @

Autogenerate scripts will translate this to liquid syntax ensuring a nice syntax highlighting on the website.

You can see the result of these changes [in a preview](http://tidal-reference.bitballoon.com/functions). The reference chapter is _mostly_ generated from the source code.


lennart:
will eventually result in tidal 0.8

PR #99, opened 2016-03-08
amend parameter description and add beat lock

telephon:
just a few details ...


telephon:
ups, I added the beat lock thing to the branch, forgetting that it hadn't been merged yet. So I just made the title of the pull request reflect both. I hope that's not too bad. If it is, just let me know and I make two pull requests.

PR #101, opened 2016-04-04
Add source-repository section to cabal file

bergmark:
This enables `cabal get -s tidal` and also gives a clickable direct link from hackage to the repo.


yaxu:
I decied to put this in 0.8-dev instead, thanks

PR #105, opened 2016-05-04
envelope parameters

bgold-cosmos:
Adds support for envelope parameters in Dirt and SuperDirt, `attack`, `hold`, and `release`.  All are times in seconds, `attack` and `release` must be specified for the envelope to take effect.


yaxu:
Merged here https://github.com/tidalcycles/Tidal/commit/ac6c74c87b3446295f332970a816e593485c69b0
Hope I got it right.. Feel like I'm flying blind with git sometimes

PR #108, opened 2016-05-20
[fix] multichannel midi

lennart:
This will rearrange things to make MIDI scheduling work correctly with multiple connections to a single or multiple channels of a MIDI device.

I tested this locally but your mileage may vary, please test before this gets merged.

MIDI and in particular PortMIDI will only work correctly if all Messages are sent with strictly increasing timestamps.

Therefore parallel running streams (e.g. m1,m2,m3)  for the same output have to be multiplexed and ordered to avoid glitches.

Changes to make this work are:

- No additional loops for MIDI scheduling
- All Streams connected to the same Output will buffer Events on each tick
- After a Stream has "ticked" the Backend is "flushed"
- The MIDI backend keeps track of when all Streams are ticked and all events will be sent out as long as they stay within this tick (noteOff's can be outside and must be deferred)
- increased default latency to match Dirt's 0.04 (or 40ms)
- Logical Onset calculation is done within the "send loop" not within the buffering (onTick)

### DIFF
--- a/Sound/Tidal/Dirt.hs
+++ b/Sound/Tidal/Dirt.hs
@@ -69,7 +69,7 @@ superDirtSlang = dirtSlang { timestamp = BundleStamp, path = "/play2", namedPara
 
 superDirtBackend port = do
   s <- makeConnection "127.0.0.1" port superDirtSlang
-  return $ Backend s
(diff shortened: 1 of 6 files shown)


lennart:
will close, since midi now lives in tidal-midi again...

PR #114, opened 2016-05-26
[fix] for `superwash`

lennart:
make `superwash` return to eventually playing the given pattern normally. This adds another parameter to define the number of cycles the fadeout function is applied.

this is not backwards compatible. jumpIn and jumpIn' have been changed accordingly.


bgold-cosmos:
Thanks for doing this, it's exactly what I was thinking of.

PR #121, opened 2016-06-02
pitch model parameters for supercollider

telephon:
fine for beyond 0.8


telephon:
not sure where this other mess comes from. git riddles!

PR #132, opened 2016-08-17
Replace random number generator - from mersenne-random-pure64 to tf-random

stephentetley:
Update - please hold off this pull request for the time being, see message below.

--------------
Hi Alex and all

Here is a pull request to change the random number generator to tf-random which is easier to compile on Windows now that the Platform no longer ships old-time (which the previous random generator mersenne-random-pure64 depends on).

The patch replicates the function _randomDouble_ (from mersenne) as tf-random only generates Word32's - via the function _next_ in the module TF.Gen.

I've tested that the new implementation of _randomDouble_ generates a comparable distribution of numbers between 0 and 1.0 to the implementation in mersenne. A zip of the test functions is attached.

Please note - both mersenne and tf-random are pseudo random number generators - they will generate the same sequences of random numbers if run with the same initial seed. Obviously the sequences mersenne generates are different to those generated by tf-random.

[test.zip](https://github.com/tidalcycles/Tidal/files/423638/test.zip)

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -12,7 +12,9 @@ import Data.Ratio
 import Debug.Trace
 import Data.Typeable
 import Data.Function
-import System.Random.Mersenne.Pure64
+import Data.Word
+import System.Random.TF
+import qualified System.Random.TF.Gen as TFGen
 import Data.Char
 import qualified Data.Text as T
 
@@ -627,7 +629,19 @@ d1 $ jux (|+| ((1024 <~) $ gain rand)) $ sound "sn sn ~ sn" # gain rand
 rand :: Pattern Double
 rand = Pattern $ \a -> [(a, a, timeToRand $ (midPoint a))]
 
-timeToRand t = fst $ randomDouble $ pureMT $ floor $ (*1000000) t
+timeToRand t = fst $ randomDouble $ mkTFGen $ floor $ (*1000000) t
+
+-- Replicate Double generation for the @tf-random@ library which only generates Word32's.
+-- Upto Tidal version 0.8 we depended on @mersenne-random-pure64@ which natively generated Doubles. 
+randomDouble :: TFGen -> (Double, TFGen)
+randomDouble g = 
+    let (n,g1) = TFGen.next g 
+        i = expandW32 n
+    in (fromIntegral (i `div` 2048)/ 9007199254740992,g1)
+
+expandW32 :: Word32 -> Word64
+expandW32 n = let i = fromIntegral n  in i*i
+
 
 {- | Just like `rand` but for integers, `irand n` generates a pattern of (pseudo-)random integers between `0` to `n-1` inclusive. Notably used to pick a random
 samples from a folder:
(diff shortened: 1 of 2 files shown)


stephentetley:
Whoops - seems like I've made a mistake.

Please hold off this request whilst I check what was gone wrong, GitHub is showing 9 files changed whereas I've only intentionally changed Pattern.hs and tidal.cabal.

yaxu:
I'm forever a git noob but I think this might be normal, e.g. could be a case of you merging in upstream changes and this somehow finding its way back in your PR (where they then have no effect)..

stephentetley:
Thanks Alex - the pull request seems to be on master rather than 0.9-dev, though.

stephentetley:
It's a bit late tonight, but I'll go through the pull request again tomorrow evening and see what I did wrong.

lennart:
@stephentetley as suggesteg, you added your feature against 0.9-dev but you are currently _wanting_ merge into master, this is where the commits are coming from. just change the pr to merge against 0.9-dev and it should be fine.

stephentetley:
Thanks Lennart - I'm unexpectedly busy this evening and have a full weekend so I won't have time to fix it til into next week. 

Sorry about the mess.

stephentetley:
As per Tom's comment on the Tidal mailing list- I've contacted Ben Gamari to see about patching mersenne-random-pure64 itself, rather than switching dependencies in Tidal. This would be preferred as it would just need a version bump to Tidal's cabal file.

stephentetley:
Hi everyone - I've changed the pull request to be against 0.9-dev. 

The change-set looks like it should - a one line (one dep) change to tidal.cabal and a small change to Pattern.hs changing imports at the top and adding the functions randomDouble and it's helper expandW32, plus slightly changing timeToRand.

expandW32 could be inlined in randomDouble, but it would need the GHC extension ScopedTypeVariables so it is top level function to ensure the multiplication of i*i is performed on a Word64.

It turns out GitHub has a new feature to handle changing which branch a pull request is made on, so it was simply a matter of editing the title and selecting the correct branch through the web interface.

I haven't heard from Ben Gamari yet concerning the alternative option - patching mersenne-random-pure64 rather than Tidal, apparently he is moving back to America from a stay in Europe.

stephentetley:
Hi everyone

Ben Gamari has updated mersenne-random-pure64 to depend on time rather than old-time (version 0.2.1.0). 
http://hackage.haskell.org/package/mersenne-random-pure64

Alex - I'll leave it up to you if you want to put a lower bounds on mersenne-random-pure64 in tidal.cabal. I think cabal should pick the latest version it can anyway.

I'll close this pull request as "not needed" as long as someone is maintaining mersenne-random-pure64 I think it makes sense to stick with it rather than change.

yaxu:
This has broken things for me.. mersenne-random-pure64 uses `diffTimeToPicoseconds` which is only available after from version 1.6 of the time package, which isn't installable in version 7.6.3 of ghc that comes with the linux mint I'm using here. I guess this isn't a problem for people with more up to date ghc installs.. A workaround is to install the old version with `cabal install mersenne-random-pure64-0.2.0.5` before installing tidal.

stephentetley:
Hi Alex - do you want me to look at putting a conditional switch in the .cabal file?

Thanks

yaxu:
Yes that'd be great!

yaxu:
Thanks @stephentetley and @lennart, will close again

Issue #135, opened 2016-08-26
`cabal install tidal` failure


chickencoder:
When I run `cabal install tidal` I get the following error... 

```
Resolving dependencies...
Configuring mersenne-random-pure64-0.2.1.0...
Building mersenne-random-pure64-0.2.1.0...
Failed to install mersenne-random-pure64-0.2.1.0
Build log ( /home/jesse/.cabal/logs/mersenne-random-pure64-0.2.1.0.log ):
Configuring mersenne-random-pure64-0.2.1.0...
Building mersenne-random-pure64-0.2.1.0...
Preprocessing library mersenne-random-pure64-0.2.1.0...
[1 of 4] Compiling System.Random.Mersenne.Pure64.Base ( dist/build/System/Random/Mersenne/Pure64/Base.hs, dist/build/System/Random/Mersenne/Pure64/Base.o )
[2 of 4] Compiling System.Random.Mersenne.Pure64.Internal ( System/Random/Mersenne/Pure64/Internal.hs, dist/build/System/Random/Mersenne/Pure64/Internal.o )
[3 of 4] Compiling System.Random.Mersenne.Pure64.MTBlock ( System/Random/Mersenne/Pure64/MTBlock.hs, dist/build/System/Random/Mersenne/Pure64/MTBlock.o )
[4 of 4] Compiling System.Random.Mersenne.Pure64 ( System/Random/Mersenne/Pure64.hs, dist/build/System/Random/Mersenne/Pure64.o )

System/Random/Mersenne/Pure64.hs:64:50:
    Not in scope: ‘diffTimeToPicoseconds’
cabal: Error: some packages failed to install:
mersenne-random-pure64-0.2.1.0 failed during the building phase. The exception
was:
ExitFailure 1
tidal-0.8 depends on mersenne-random-pure64-0.2.1.0 which failed to install.
```

My system is Ubuntu 15.10, thanks

chickencoder:
Solved by installing the previous version of mersenne-random-pure64 (0.2.0.5)

lennart:
relates to #147

lennart:
closing, feel free to reopen

PR #136, opened 2016-08-27
[fix] force compatible mersenne-random version

lennart:
make sure new users with ghc below 7.8 can still install tidal, as mersenne random pure package (0.2.1.0) will break.

this might need to be accompained by a version bump to 0.8.1 and published to hackage


lennart:
relates to #135 and #132

PR #139, opened 2016-08-31
Fix websocket connection Eq instance (caused infinite loop)

vivid-synth:
The WS.Connection Eq instance didn't do anything -- whenever it was called it'd infinitely loop: neither ```==``` nor ```/=``` were defined.

As an example of what a standalone ```instance Eq``` does:

```
data A = A

instance Eq A

main = print $ A == A
```

```main``` loops forever, never returning.

I created a new type because we can't make one on ```WS.Connection``` itself.


vivid-synth:
Fixed second conflict here.

PR #141, opened 2016-08-31
Remove incoherent and overlapping instance pragmas?

vivid-synth:
(This one I'm not positive about:)

I wrote a patch to move the ```OVERLAPPING``` / ```INCOHERENT``` pragmas to the instance sites, but are they needed at all? I'm able to use Tidal fine without them. ```INCOHERENT``` is a pretty nasty one in corner cases so it'd be great to remove if we can.


yaxu:
Thanks a lot for the PRs! Would it be possible to resubmit them to the `0.9-dev` branch please?

vivid-synth:
@yaxu done!

yaxu:
Thanks a lot!

PR #142, opened 2016-09-02
[add] .travis.yml continuous integration config

lennart:
mergable when travis build passes...


lennart:
also fixes compatibility with 7.6 and older cabal versions

PR #145, opened 2016-09-13
Add intermediate parsing AST

vivid-synth:
Instead of parsing straight from ```String``` to ```Pattern``` (and, since ```Pattern```s are functions, we're never able to inspect/manipulate the parsed ```Pattern```s directly), we parse to an AST which we then immediately convert to a ```Pattern```. Everything should behave identically, but we'll have more flexibility.


lennart:
this might be really helpful to implement rotation `(~>)` within the dsl as @bgold-cosmos proposed!

One question regarding namespacing, do we really need the `TPat_` prefix for all the constructors? Can one not solve namespace issues on import with `import qualified ... as ...`?

yaxu:
This looks like it could simplify things nicely! I'd be a little bit nervous about putting too many new features into the parser but let's see..

PR #148, opened 2016-10-07
Some small cleanup

vivid-synth:
Small fix


vivid-synth:
Whoops, 2 different commits in here. The goal here though is clean code for both - maybe I don't need to split them?

lennart:
@vivid-synth these corrections are great and as far as I can see good to merge, however github tells us otherwise, can you update your branch to latest 0.9-dev?

PR #149, opened 2016-10-07
Simplify env var lookup

vivid-synth:
This way we don't need to throw an exception

### DIFF
--- a/Sound/Tidal/Tempo.hs
+++ b/Sound/Tidal/Tempo.hs
@@ -9,7 +9,7 @@ import Control.Monad (forM_, forever, void)
 import Control.Concurrent (forkIO, threadDelay)
 import Control.Concurrent.MVar
 import Control.Monad.Trans (liftIO)
-import Data.Maybe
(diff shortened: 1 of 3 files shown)


Sound/Tidal/Tempo.hs
```diff
@@ -9,15 +9,17 @@ import Control.Monad (forM_, forever, void)
 import Control.Concurrent (forkIO, threadDelay)
 import Control.Concurrent.MVar
 import Control.Monad.Trans (liftIO)
+import Data.Maybe
```
lennart:
```haskell
import Data.Maybe (fromMaybe,maybe)
```
 ?

vivid-synth: K!

lennart:
relates to #157

lennart:
@vivid-synth if you update to latest 0.9-dev once again (I just merged the update to cabal 1.18) travis should pass!

vivid-synth:
Updated

lennart:
great, should be good to merge.

PR #152, opened 2016-10-27
tidal: set feedback parameter to zero by default

telephon:
Not sure if it is still necessary, but here it is.

### DIFF
--- a/Sound/Tidal/Params.hs
+++ b/Sound/Tidal/Params.hs
@@ -114,7 +114,7 @@ Using `cut "0"` is effectively _no_ cutgroup.
 -- | a pattern of numbers from 0 to 1. Sets the level of the delay signal.
 (delay, delay_p)                 = pF "delay" (Just 0)
 -- | a pattern of numbers from 0 to 1. Sets the amount of delay feedback.
-(delayfeedback, delayfeedback_p) = pF "delayfeedback" (Just (-1))
+(delayfeedback, delayfeedback_p) = pF "delayfeedback" (Just 0)
 -- | a pattern of numbers from 0 to 1. Sets the length of the delay.
 (delaytime, delaytime_p)         = pF "delaytime" (Just (-1))
 (detune, detune_p)               = pF "detune" (Just 0)


telephon:
@lennart All in all, wouldn't it be better to change them all to `Nothing` and then keep the defaults in classic dirt and superdirt respectively?

lennart:
as of now, specifying a param with default `Nothing` will make these mandatory. which is right now just `s`.

I don't know why there is this restriction.

bgold-cosmos:
Dirt doesn't receive messages the same way SuperDirt does - it expects OSC messages where all parameters are in a certain order and have values.  In fact, I think this is the only reason for the "defaults" in Param.hs these days, my understanding is that SuperDirt will never sees those defaults.

My recollection is that `-1` defaults were a way of working around whether a parameter had been used in Tidal or not, equivalent to how SuperDirt can test for `nil`.

telephon:
ok, so we can actually jut leave it as it is? Should I close?

Sound/Tidal/Params.hs
```diff
@@ -114,7 +114,7 @@ Using `cut "0"` is effectively _no_ cutgroup.
 -- | a pattern of numbers from 0 to 1. Sets the level of the delay signal.
 (delay, delay_p)                 = pF "delay" (Just 0)
 -- | a pattern of numbers from 0 to 1. Sets the amount of delay feedback.
-(delayfeedback, delayfeedback_p) = pF "delayfeedback" (Just (-1))
+(delayfeedback, delayfeedback_p) = pF "delayfeedback" (Just 0)
```
lennart: just read code again, we cannot change defaults without changing classic dirt.

lennart:
currently we cannot change this without changing classic dirt as well, do we **want** to change classic dirt as well?

telephon:
@lennart we want no feedback unless specified. This currently works on both dirts, I suppose, doesn't it?

yaxu:
@telephon the problem is that delay is a global effect. If pattern a doesn't set `delayfeedback`, it would turn off feedback for a sound made with pattern b.
Could patch classic dirt to ignore `delayfeedback` if `delay` is set to 0, but why change this if superdirt ignores the default anyway?

PR #156, opened 2016-12-02
0.9 extrautils

bgold-cosmos:
This is a combination of the small utility functions that I've been using the most, some of which have been picked up by other people.  The individual commits have short descriptions (and I've tried to include comments in the code that can be auto-harvested), but here's a short index:

  * swing, swingBy - alter pattern timing
  * toScale - select from notes in a scale
  * every' - every with an offset
  * scalex - exponential version of scale
  * somecycles, somecyclesBy - like sometimes, but per cycle rather than per event
  * (***) (+++) (###) - pattern merging operators on lists, can save some typing

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -340,6 +340,10 @@ every :: Int -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
 every 0 _ p = p
 every n f p = when ((== 0) . (`mod` n)) f p
 
+-- | @every n o f'@ is like @every n f@ with an offset of @o@ cycles
+every' :: Int -> Int -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
+every' n o f = when ((== o) . (`mod` n)) f
+
 -- | @foldEvery ns f p@ applies the function @f@ to @p@, and is applied for
 -- each cycle in @ns@.
 foldEvery :: [Int] -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -1170,3 +1170,12 @@ toScale' o s p = (+) <$> fmap (s!!) notep <*> fmap (o*) octp
 
 toScale::[Int] -> Pattern Int -> Pattern Int
 toScale = toScale' 12
+
+{- | `swingBy x n` divides a cycle into `n` slices and delays the notes in
+the second half of each slice by `x` fraction of a slice . `swing` is an alias
+for `swingBy (1%3)`
+-}
+swingBy::Time -> Time -> Pattern a -> Pattern a 
+swingBy x n = inside n (within (0.5,1) (x ~>))
+
+swing = swingBy (1%3)
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -700,6 +700,13 @@ almostAlways = sometimesBy 0.9
 never = flip const
 always = id
 
+{- | `somecyclesBy` is a cycle-by-cycle version of sometimesBy, and has a
+`somecycles = somecyclesBy 0.5` alias -}
+somecyclesBy x = when (test x)
+  where test x c = (timeToRand $ fromIntegral c) < x
+
+somecycles = somecyclesBy 0.5
+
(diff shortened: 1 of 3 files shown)


Sound/Tidal/Stream.hs
```diff
@@ -218,6 +218,11 @@ infixl 1 |/|
 (|/|) :: ParamPattern -> ParamPattern -> ParamPattern
 (|/|) = mergeNumWith (div) (/)
 
+(###) = foldl (#)
```
lennart: although it might seem obvious for a haskell programmer what these do,
can you add example usages or short explanations as haddock comments for one of these at least?

still hoping we can somehow use these docs somewhere else as well....

bgold-cosmos: added!

Sound/Tidal/Stream.hs
```diff
@@ -218,6 +218,17 @@ infixl 1 |/|
 (|/|) :: ParamPattern -> ParamPattern -> ParamPattern
 (|/|) = mergeNumWith (div) (/)
 
+{- | These are shorthand for merging lists of patterns with @#@, @|*|@, @|+|@,
```
telephon: questions only: wouldn't it be more general to add a very short definition of `foldl` which you can combine with any function? something like `(^ #)` Well, I see that this doesn't type itself so well. 

Or, alternatively, why not only `##` but three of them?

lennart: just guessing but maybe one of: `##`, `//`, `++` gotcha may already be used in standard haskell

vivid-synth: Just a note: you can see if an operator has been "taken" in haskell by searching it on Hayoo (http://hayoo.fh-wedel.de/) -- for operators, enclose them in parens like ```(++)```

bgold-cosmos: There might be a higher-level way of defining it more simply, but I couldn't figure it out.

I used three across the board for consistency, and because `**` is already used for exponentiation and `++` for list concatenation.

Really, these only exist because I got tired of typing long compositions of sections, like
`every 2 ((|*| speed "2") . (|*| end "0.5") . (|*| shape "2"))`, and this was the only thing I could come up with that avoided paren-matching hell and didn't involve lots of extra keystrokes.

lennart: [(###)](http://hayoo.fh-wedel.de/?query=%28%23%23%23%29)
[(+++)](http://hayoo.fh-wedel.de/?query=%28%2B%2B%2B%29)
[(///)](http://hayoo.fh-wedel.de/?query=%28%2F%2F%2F%29)
[(***)](http://hayoo.fh-wedel.de/?query=%28***%29)

lennart: I think we can go with these for now.

Just tried them for a few hours and the improvement from having _any_ operator to work on a a list of `ParamPattern` is huge!

These however are especially handy since they are easily memorable.

We can change to different _names_ later on, but I'd rather have these shortcuts **in** 0.9 than not.

What would be interesting to know is, whether these _helpers_ actually improve usability for others or rather add confusion (even more symbols!)

Maybe this however depends on the prominence of these shortcuts within our docs (right now this would only show on hackage, however, I just read, people get their information about how Tidal works from exactly there!)

Sound/Tidal/Pattern.hs
```diff
@@ -700,6 +704,13 @@ almostAlways = sometimesBy 0.9
 never = flip const
 always = id
 
+{- | `somecyclesBy` is a cycle-by-cycle version of sometimesBy, and has a
```
lennart: replace backticks with `@` for highlight (and I think referencing) of functions. also `sometimesBy` should be marked I guess.

Sound/Tidal/Pattern.hs
```diff
@@ -1145,3 +1156,26 @@ seqPLoop :: [(Time, Time, Pattern a)] -> Pattern a
 seqPLoop ps = timeLoop (maxT - minT) $ minT <~ seqP ps
   where minT = minimum $ map fst' ps
         maxT = maximum $ map snd' ps
+
+{- | `toScale` lets you turn a pattern of notes within a scale (expressed as a
```
lennart: same as `somecyclesBy`

lennart:
@bgold-cosmos thanks! think we're done, good to merge?

bgold-cosmos:
No objections here

PR #157, opened 2016-12-05
[++] cabal >= 1.18 in build matrix

lennart:
our dependency graph includes `hmt` which depends on `safe` which requires cabal 1.18 from 0.3.10 (released Nov. 8th 2016 ) on.

right now this pull request only enforces the cabal version in travis continous integration builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_cache:
 
 matrix:
   include:
-    - env: CABALVER=1.16 GHCVER=7.6.3
+    - env: CABALVER=1.18 GHCVER=7.6.3
       compiler: ": #GHC 7.6.3"
       addons: {apt: {packages: [cabal-install-1.16,ghc-7.6.3], sources: [hvr-ghc]}}
     - env: CABALVER=1.18 GHCVER=7.8.4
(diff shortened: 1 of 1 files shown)


lennart:
do I miss anything, or does this change imply changing `cabal-version: >=1.18` in `tidal.cabal` as well?

if so, it might be good, to again release a newer version, i.e. 0.8.3, no?

vivid-synth:
Tbh I don't think hmt is actually used in tidal anymore. It was attempted to be removed twice:

> commit ebe4f56fddaf4d8657c60b72ce19a3f1262a1d84
> Author: alex <alex@slab.ogr>
> Date:   Thu Aug 25 20:02:56 2016 +0100
> 
>     remove hmt, process, transformers, bytestring deps..


> commit 2fd896e9bde05d7e369e6cc8b81dcdc5e37a262a
> Author: alex <alex@slab.ogr>
> Date:   Thu Aug 25 19:49:53 2016 +0100
> 
>     remove hmt dependency (until we need it more)

I don't see anything grepping:

`grep -R 'import Music.Theory'`

lennart:
@vivid-synth that's right, but this is on master and the changes are introduced in 0.9-dev

however one of your other pull requests will eventually add an explicit dependency for `safe` and therefore break anyhow (if we don't bump cabal to >=1.18)

lennart:
this will have to also be done for 0.9-dev to make builds pass...

PR #160, opened 2016-12-29
parameters: add multichannel panning parameters

telephon:
This corresponds to the branch with the same name in SuperDirt: https://github.com/musikinformatik/SuperDirt/pull/40.

Will probably have some changes before merging.


fdragovic:
I´m confused, how does the stereo width parameter work when there is already a spread function in tidal? (https://hackage.haskell.org/package/tidal-0.8.2/docs/src/Sound-Tidal-Pattern.html#spread)

bgold-cosmos:
The `spread` function actually has nothing to do with stereo panning, it "spreads" a function that takes a numerical parameter over a whole list of numerical parameters.

lennart:
@fdragovic it won't even compile like this: https://travis-ci.org/tidalcycles/Tidal/jobs/187550015

we will have to rethink naming for _stereo_ spread, since I guess a lot users are used to `spread` the function

fdragovic:
when I saw the discussion I thought it was sc specific so I didnt comment, but I was thinking width should be free, and is often used to label stereo spread anyway? (:

telephon:
Ah sure, yes, we can rename `spread` in multichannel audio. I forgot that the parameter names in Haskell are also bound symbols.

telephon:
Maybe `span`?

telephon:
works now?

PR #163, opened 2017-01-12
shuffle and scramble

bgold-cosmos:
`shuffle` makes random permutations of Patterns, `scramble` randomly select elements from
Patterns.  Also includes `cycleChoose` as a helper function (which might be useful on its own).

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -1181,3 +1181,25 @@ swingBy::Time -> Time -> Pattern a -> Pattern a
 swingBy x n = inside n (within (0.5,1) (x ~>))
 
 swing = swingBy (1%3)
+
+{- | `cycleChoose` is like `choose` but only picks a new item from the list
+once each cycle -}
+cycleChoose::[a] -> Pattern a
+cycleChoose xs = Pattern $ \(s,e) -> [((s,e),(s,e), xs!!(floor $ (dlen xs)*(ctrand s) ))]
+  where dlen xs = fromIntegral $ length xs
+        ctrand s = timeToRand $ fromIntegral $ floor $ sam s
+
+{- | `shuffle n p` divides one cycle of the `p` into `n` parts and returns a
+new random permutation each cycle.  For example, `shuffle 4 "bd sn cp hh"` 
+might return `"sn cp bd hh"` one cycle and `"hh cp bd sn"` the next.
+-}
+shuffle::Int -> Pattern a -> Pattern a
+shuffle n = fit' 1 n (run n) (unwrap $ cycleChoose $ map listToPat $ 
+  permutations [0..n-1])
+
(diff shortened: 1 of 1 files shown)


Sound/Tidal/Pattern.hs
```diff
@@ -1181,3 +1181,25 @@ swingBy::Time -> Time -> Pattern a -> Pattern a
...
+
+{- | `cycleChoose` is like `choose` but only picks a new item from the list
+once each cycle -}
+cycleChoose::[a] -> Pattern a
+cycleChoose xs = Pattern $ \(s,e) -> [((s,e),(s,e), xs!!(floor $ (dlen xs)*(ctrand s) ))]
+  where dlen xs = fromIntegral $ length xs
+        ctrand s = timeToRand $ fromIntegral $ floor $ sam s
+
+{- | `shuffle n p` divides one cycle of the `p` into `n` parts and returns a
+new random permutation each cycle.  For example, `shuffle 4 "bd sn cp hh"`
```
telephon: what does "permutation" mean here? Permutation of n parts? The example doesn't really show how it works (or I just don't get it!).

bgold-cosmos: Yes, basically it turns the pattern into a list of n parts, then rearranges them randomly back into a pattern.  I'll see if I can clarify the example a bit.

telephon: ah so they are not the actual permutations, but random selections? So from a pattern `bd hh cp`` you could get the sequence `bd bd hh`, e.g. or are only `bd cp hh`, `cp bd hh` and `cp hh bd` possible?

bgold-cosmos: No, they **are** permutations - that's the difference between `shuffle` (permutations) and `scramble` (random selections).  I think I'll just write out a lengthy example for clarity, it's not like space is at a premium.

bgold-cosmos: Yes, maybe the new documentation is clearer?

Sound/Tidal/Pattern.hs
```diff
@@ -1189,16 +1189,20 @@ cycleChoose xs = Pattern $ \(s,e) -> [((s,e),(s,e), xs!!(floor $ (dlen xs)*(ctra
   where dlen xs = fromIntegral $ length xs
         ctrand s = timeToRand $ fromIntegral $ floor $ sam s
 
-{- | `shuffle n p` divides one cycle of the `p` into `n` parts and returns a
-new random permutation each cycle.  For example, `shuffle 4 "bd sn cp hh"` 
-might return `"sn cp bd hh"` one cycle and `"hh cp bd sn"` the next.
+{- | `shuffle n p` evenly divides one cycle of the pattern `p` into `n` parts,
+and returns a random permutation of the parts each cycle.  For example,
```
telephon: now it is very clear. Maybe "returns a randomly chosen permutation" ?

PR #164, opened 2017-01-23
drum now uses `midinote` instead of `n`

lvm:
Small fix for (percussion) midi devices


lvm:
Closing this and reopening for the `0.9dev` branch

PR #165, opened 2017-01-24
Params for a possible phaser DSP.

pd3v:


telephon:
There are some unrelated changes here - tremolo should already be in here. Could you also here just try and make a fresh PR? Thanks!!

PR #180, opened 2017-03-06
increase latency default to avoid late OSC  messages

telephon:


telephon:
would it be easy to make this value settable from tidal? I'd be happy to make this shorter if one could adjust it if necessary.

PR #195, opened 2017-04-16
Port the monoid fix

yaxu:


Sound/Tidal/Pattern.hs
```diff
@@ -4,7 +4,7 @@
 module Sound.Tidal.Pattern where
 
 import Control.Applicative
--- import Data.Monoid
+import Data.Monoid (mappend, mempty, Monoid())
```
vivid-synth: `Monoid()` here is unecessary - the class is imported as soon as you import the module

vivid-synth: On second thought, these are all not needed: you can simply write `import Data.Monoid ()`

PR #216, opened 2017-05-28
use Num for scales and chords

bgold-cosmos:
This allows for more flexible use of scales and the `toScale` function,
for example
`d1 $ s "jvbass" # up (toScale Scales.major $ choose [0..12])`


yaxu:
Sorry didn't see this before! A good idea.

PR #218, opened 2017-06-13
emacs25 compat

### LINKED ISSUE #198: Unmark region after C-return in spacemacs
Hi. I'm using the provided `tidal.el` package with [spacemacs](http://spacemacs.org/) - a vim like editor that is based on emacs.
The main issue with the package is that after `C-return` the paragraph remain in visual mode, and I need to hit `escape` to return to select mode.
The solution I found is to add `(deactivate-mark)` at the end of `tidal-run-multiple-lines`.
I'm far from knowing how to properly work with emacs lisp, and it might be a problematic solution for most (emacs) users. I really don't know...
However, would you consider adding something like this to the package? Is there a benefit from something like this for someone that doen't use spacemacs?

lennart:
this will use emacs `save-mark-and-excursion` in version 25+ and should fix #198 

however, I have not tested this against 24 if it will still work there! Would be happy if someone could check backwards compatibility.

### DIFF
--- a/tidal.el
+++ b/tidal.el
@@ -187,7 +187,7 @@
   (if (>= emacs-major-version 25)
       (save-mark-and-excursion
        (tidal-eval-multiple-lines))
-    (save-mark
+    (save-excursion
      (tidal-eval-multiple-lines))
     )
   )
--- a/tidal.el
+++ b/tidal.el
@@ -164,25 +164,32 @@
   (next-line)
   )
 
+(defun tidal-eval-multiple-lines ()
+  "Eval the current region in the interpreter as a single line."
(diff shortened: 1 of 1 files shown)


yaxu:
I just tried it in emacs24 and got this message when trying to evaluate a block (via ctrl-enter):
`if: Symbol's function definition is void: save-mark`

Single line evaluation worked, though.

yaxu:
Did you mean to put `save-excursion` ?

yaxu:
Yes switching `save-mark` for `save-excursion` fixes it for emacs24. Have committed the change.

lennart:
ok looks good to merge.

Issue #133, opened 2016-08-18
tidal streams should process at a fixed rate per second


yaxu:
Currently do it at a fixed rate per cycle, which means if cps is low, it takes a long time for changes to patterns to take effect.
It would still be nice if changes could be quantised, but this could be done with a transition function.

telephon:
In sc, each node- or pattern-proxy has variables that allow you to specify when (and to a degree also how) the transition to the replacing code happens.

lennart:
I hope no one considers the current behavior _correct_, so flagging this as a bug.

lennart:
thinking about this:

1. cps bound
tick patterns x times per cycle
pro: changes to pattern at end of a cycle will take effect on the *next* cycle
cons: time for evaled patterns to take effect could literally take ages!

2. unbound, fixed
tick patterns at fixed frequency
pro: same time for evaled patterns to take effect, regardless of perceived "tempo"
cons: might lead to unwanted glitches since eval frequency might not be in time with tidal clock

So what we want is:

- changes just before end of cycle take effect on *next* cycle
- changes never take a maximum amount of time

?

bgold-cosmos:
Would it solve both issues to check x times per cycle, where `x` is something like `floor (8/cps)`?  Any weird glitches that could cause?

yaxu:
If I understand you right that's the current arrangement @bgold-cosmos, with the current setting of 4 per cycle.

@lennart yes I agree in my case but others might have different needs, an alternative way of looking at it is you either want
1/ evaluation/change on the sam (cycle boundary)
2/ evaluation/change as soon as possible

Want you want might change depending on the situation.

bgold-cosmos:
Currently it's fixed at 4 (or 8?) per cycle and doesn't depend on cps; I was thinking it could just switch dynamically to more ticks per cycle if cps is changed to something very low.

yaxu:
Gotcha, yes that makes sense, maybe checking once per cycle is safest?

d0kt0r0:
Hmmm... When patterns execute over more than one cycle, for example as a result of using "slow", the same issue recurs. Once per cycle with a pattern that has had slow 8 applied to it will be redefined at the next 1/8 of the way through the pattern.

vivid-synth:
An idea:

At 60 cps it's 4x per cycle

```
|-----|-----|-----|-----| 4 x 250ms
```

At 30 cps, we want still to be updating at 250ms
```
[...]

PR #222, opened 2017-08-03
Espgrid

d0kt0r0:
adds EspGrid and Multimode modules for controlling sync. Additional dependency on control-monad-loops. Continues to work in the normal way without EspGrid, etc.


Sound/Tidal/EspGrid.hs
```diff
@@ -0,0 +1,114 @@
...
+
+import Sound.Tidal.Tempo
+import Sound.Tidal.Time as T
+import Sound.Tidal.Stream
+import Sound.Tidal.Dirt
+import Sound.Tidal.Transition (transition)
+import Sound.Tidal.Pattern (silence)
+
+parseEspTempo :: [Datum] -> Maybe Tempo
+parseEspTempo d = do
```
vivid-synth: It's probably more idiomatic to say ```parseEspTempo [on', bpm', t1', t2', n'] = do```, and then referring to the args by name below instead of using ```(!!)```

tidal.cabal
```diff
@@ -53,6 +55,7 @@ library
     , safe
     , websockets > 0.8
     , mtl >= 2.1
+    , monad-loops
```
vivid-synth: This is a very small dependency; I'd be +1 to adding it

Sound/Tidal/Tempo.hs
```diff
@@ -154,39 +154,9 @@ bpsUtils = cpsUtils
...
-          do now <- getCurrentTime
-             let delta = realToFrac $ diffUTCTime now (at t)
-                 actualBeat = (beat t) + ((cps t) * delta)
-                 beatDelta = (fromIntegral b) - actualBeat
-                 delay = beatDelta / (cps t)
-             threadDelay $ floor (delay * 1000000)
-             callback t b
-             return $ b + 1
--}
-
```
vivid-synth: @yaxu I've assumed some of this commented-out stuff was intentionally left in; do we want to remove it?

Sound/Tidal/EspGrid.hs
```diff
@@ -0,0 +1,114 @@
...
+  on <- datum_integral (d!!0)
+  bpm <- datum_floating (d!!1)
+  t1 <- datum_integral (d!!2)
+  t2 <- datum_integral (d!!3)
+  n <- datum_integral (d!!4)
+  let nanos = (t1*1000000000) + t2
```
vivid-synth: Nitpick: I might say ```(10^9)``` or ```(10**9)``` instead of ```1000000000```, for readability

Sound/Tidal/EspGrid.hs
```diff
@@ -0,0 +1,114 @@
...
+  let utc = posixSecondsToUTCTime ((realToFrac nanos)/1000000000)
```
vivid-synth: Same as above

Sound/Tidal/EspGrid.hs
d0kt0r0: One difference though would be that with the "one by one" deindexing it would work were the message/list to gain extra data, whereas with the pattern matching it is tolerant to the addition of extra information to the message. I don't have strong opinions about it, but it is not just a question of idiom...

Sound/Tidal/EspGrid.hs
```diff
@@ -0,0 +1,114 @@
...
+  (mTempo, _) <- runClientEsp
+  nowBeat <- getCurrentBeat mTempo
+  let nextTick = ceiling (nowBeat * (fromIntegral tpb))
+  iterateM_ (clockedTickLoopEsp tpb callback mTempo) nextTick
+
+clockedTickLoopEsp :: Int -> (Tempo -> Int -> IO ()) -> MVar Tempo -> Int -> IO Int
+clockedTickLoopEsp tpb callback mTempo tick = do
+  tempo <- readMVar mTempo
+  if (paused tempo)
+    then do  -- TODO - do this via blocking read on the mvar somehow rather than polling
```
vivid-synth: For a blocking read you might want to use a ```TVar``` instead of ```MVar```, and something like ```check``` or ```retry``` to check if ```paused``` has changed (http://hackage.haskell.org/package/stm-2.4.4.1/docs/Control-Monad-STM.html)

Sound/Tidal/EspGrid.hs
d0kt0r0: (sorry that should have said "with the pattern matching it is *not* tolerant to the addition of extra information)

Sound/Tidal/EspGrid.hs
d0kt0r0: Yeah... and/but I think that comment about the blocking read is just copied negligently from elsewhere in the Tidal source because I actually don't see any reason for this to be anything but a non-blocking read. One wouldn't want the loop rendering sound events to be held up by the processing of an incoming networking packet, I would tend to think!

Sound/Tidal/EspGrid.hs
d0kt0r0: That's helpful - thank you!

Sound/Tidal/EspGrid.hs
vivid-synth: (A quick sketch of how you'd use it):

```
 .  :m + Control.Concurrent
 .  :m + Control.Concurrent.STM
 .  :m + Control.Monad
 .  t <- newTVarIO (5 :: Int)
 .  forkIO $ forever $ do { atomically (do { x <- readTVar t ; check (x > 10) ; writeTVar t 0}) ; putStrLn "over 10!" }
ThreadId 797
 .
 .  atomically $ writeTVar t 5
 .  atomically $ writeTVar t 11
over 10!
 .  atomically $ writeTVar t 3
 .  atomically $ writeTVar t 13
over 10!
```

vivid-synth:
I've added some comments to the code; not sure why they're not showing up here.

Sound/Tidal/EspGrid.hs
vivid-synth: True! If you wanted to support extra args you'd probably say

```parseEspTempo (on' : bpm' : t1' : t2' : n' : _) = do```

The reasons for this style are:
  - Easier not to mix up arguments and numbers, and you can just call them by their names
  - ```(!!)``` is somewhat frowned upon, as it's a partial function. The pattern-matching is partial too, but you get a warning from the compiler about it

Sound/Tidal/EspGrid.hs
vivid-synth: I think this process is forked, so it wouldn't block any other processes.

Sound/Tidal/EspGrid.hs
d0kt0r0: Righto!

tidal.cabal
yaxu: Looks useful!

Sound/Tidal/Tempo.hs
yaxu: I'm in favour of deleting this and probably all commented out code.

yaxu:
This should target the active development branch `1.0-dev` rather than master I think

yaxu:
Rather than replicate Tempo.hs, I think it would be better for espgrid to interact with the tidal tempo server. This would allow a running tidal instance to slave to espgrid without restarts. Ideally espgrid would announce itself on startup, and tidal would then automatically slave to it.

d0kt0r0:
I'm happy to refactor in various ways going forward but can we integrate this now, all these caveats and suggestions notwithstanding? The commits here have zero impact (except for an insignificant dependency on a completely standard Haskell library) on other users of Tidal, but for people who want to use EspGrid with Tidal it removes the step of having to find/clone/build a "custom" installation of Tidal. (By the way, this code also allows a running Tidal instance to slave to EspGrid without restarts - by evaluating "esp" - you can also unslave by evaluating "nosync" - there could be a larger discussion about various configuration issues with Tidal but the least I would say here is that automatic may not always be the best thing.)

d0kt0r0:
PS - re targeting branches: I'm deliberately targeting master rather than the dev branch because this is code that has no impact on anyone except those trying to use Tidal with EspGrid, for whom, just being able to a standard installation of Tidal and have what they need is an advantage. I'd be happy for continuing work on it to happen in another branch of course, but this is ready to use and has been in use for about a year.

yaxu:
The next release to hackage/stackage will be prepared from the 1.0-dev branch, please target that.

Your approach here duplicates buggy code (https://github.com/tidalcycles/Tidal/issues/133) and takes a conflicting approach with what I'm actively developing. I think it's worth a bit extra work to get things cleaner from the start. Alternatively we could release as a separate tidal-espgrid module?

d0kt0r0:
I don't see the old ticks per cycle (8 was it?) as buggy - it's just a slightly different behaviour than the new ticks per cycle, 1 (which I hadn't even noticed using Tidal possibly because I don't use low cps values very often). Similarly, this extension doesn't conflict with anything in 0.9.4 or, as it happens, 1.0-dev (so far). I've changed the target anyway, and I'm happy to continue to track and harmonize with collective efforts, but I'm having a hard time seeing any conflict here. If I change the ticks per cycle to 1 can we merge it? Or is there another difference I'm missing?

yaxu:
Ticks per cycle isn't the issue. Tempo.hs needs work to operate at a fixed rate per second.

I don't want to hold up useful features unnecessarily but let me do a bit more work on the topic-osc-tempoclock branch over the next few days to clarify the best way forward. If I don't make progress we can merge next week and release 0.9.5.

d0kt0r0:
Alright. Thinking about the fixed rate / ticks per cycle issue has there been any consideration of something like this: ticksPerCycle = whatever power of 2 results in a tick frequency within a certain range (approximately a quarter second), with the ticks per cycle redetermined at each tick?

yaxu:
Thanks @d0kt0r0!

d0kt0r0:
P {margin-top:0;margin-bottom:0;}


Hope it still works! Probably worth another look sometime soon...

PR #213, opened 2017-05-25
0.9.5

yaxu:

PR #223, opened 2017-08-07
patternify stut' and add stutx

bgold-cosmos:
upgrades `stut'` to work with patterns for the first two arguments, and adds a new version `stutx` to make it easier to prevent overlapping repeats.  

To help this work there's also a new function `durPattern`, which returns a pattern of event times.  So `durPattern "[a ~] c"` is equivalent to `"[0.25 ~] 0.5"`


bgold-cosmos:
Any objections to merging this to 1.0-dev?  The `stut'` change will require minor changes to https://github.com/tidalcycles/Tidal/pull/213/commits/e5edede7cdacfa730f99c969460f80590f9b52a0 for the 0.9.5 release, but it should be easy to do.

Issue #220, opened 2017-07-08
Raise error if choose is given empty list


yaxu:

pd3v:
Not sure if this is what you're talking about.

```
choose :: [a] -> Pattern a
choose [] =  throw (ErrorCall "Empty list. Nothing to choose from.")  -- new line here
choose xs = (xs !!) <$> (irand $ length xs)

```
This is setting the error message, nothing fancy. I wonder if a better solution could be - in case of an empty list, keep playing  the previous working `choose` list.

yaxu:
Yes looks good. I think if an error is thrown like this, the previous pattern will continue running.

pd3v:
It stops playing. If you want me to PR this solution, please let me know.

yaxu:
yes please @pd3v

yaxu:
against the 1.0-dev branch

pd3v:
@yaxu, this one is ✅.

PR #225, opened 2017-08-16
To resolve issue #220

pd3v:
To resolve issue #220 - Raise error if choose is given empty list

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -843,7 +843,7 @@ d1 $ sound (samples "xx(3,8)" (tom $ choose ["a", "e", "g", "c"]))
 plays a melody randomly choosing one of the four notes \"a\", \"e\", \"g\", \"c\".
 -}
 choose :: [a] -> Pattern a
-choose [] =  E.throw (ErrorCall "Empty list. Nothing to choose from.")
+choose [] =  E.throw (E.ErrorCall "Empty list. Nothing to choose from.")
 choose xs = (xs !!) <$> (irand $ length xs)
 
 {- |


Sound/Tidal/Pattern.hs
```diff
@@ -842,6 +842,7 @@ d1 $ sound (samples "xx(3,8)" (tom $ choose ["a", "e", "g", "c"]))
 plays a melody randomly choosing one of the four notes \"a\", \"e\", \"g\", \"c\".
 -}
 choose :: [a] -> Pattern a
+choose [] =  throw (ErrorCall "Empty list. Nothing to choose from.")
```
pd3v: This is the actual line I've added. Don't know why the others lines appear in this PR.

bgold-cosmos:
FYI, the build is failing because `throw` and `ErrorCall` require `import Control.Exception`

Sound/Tidal/Pattern.hs
vivid-synth: @pd3v :
  - You can/should just write ```error "Empty list. Nothing to choose from."``` (no ```throw``` or ```ErrorCall```)
  - The other lines are probably appearing in the diff because your editor/IDE are trimming trailing newlines

vivid-synth:
Also, should this be erroring or just returning an "empty" pattern?

Sound/Tidal/Pattern.hs
pd3v: Hi @vivid-synth, with `error`... the returned message is a bit more unclear with its additional compiler info. For live coding is that relevant? About erroring or returning `silence`, I have not chosen silence because the issue says to "Raise error (...)", I guess to give some info of what is happening.

pd3v: and about the other lines, should I leave it that way or make a new PR?

pd3v:
Hooray! 🎉

Sound/Tidal/Pattern.hs
```diff
@@ -22,6 +22,7 @@ import Sound.Tidal.Utils
 import Sound.Tidal.Bjorklund
 
 import Text.Show.Functions ()
+import Control.Exception.throw
```
vivid-synth: This is written ```import Control.Exception (throw)```

PR #226, opened 2017-08-21
Temporal params

bgold-cosmos:
provide versions of `temporalParam` to "patternify" functions that take multiple non-pattern arguments. Also "patternify" all the `striate` variants


Sound/Tidal/Pattern.hs
```diff
@@ -493,8 +493,11 @@ _every 0 _ p = p
 _every n f p = when ((== 0) . (`mod` n)) f p
 
 -- | @every n o f'@ is like @every n f@ with an offset of @o@ cycles
-every' :: Int -> Int -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
-every' n o f = when ((== o) . (`mod` n)) f
+every' :: Pattern Int -> Pattern Int -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
```
telephon: Every `"2 3 1"`th time is a little hard to make sense of - is it like an interference pattern?

PR #228, opened 2017-08-30
Fix issue with toScale.

erwald:
It seemed to be behaving strangely with patterns of overlapping
events, or patterns containing events that spanned multiple cycles.

This version fixes that by mapping over the pattern just once,
rather than doing it once for the note value and once for the octave
and then combining the two.

Also touch up documentation.

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -1457,9 +1457,9 @@ to change this use `toScale' size`.  Example:
 -}
 toScale' :: Int -> [Int] -> Pattern Int -> Pattern Int
 toScale' o s = fmap noteInScale
-  where note x = x `mod` length scale
-        octave x = x `div` length scale
-        noteInScale x = (scale !!! note x) + o * octave x
+  where note x = x `mod` length s
+        octave x = x `div` length s
+        noteInScale x = (s !!! note x) + o * octave x
 
 toScale :: [Int] -> Pattern Int -> Pattern Int
 toScale = toScale' 12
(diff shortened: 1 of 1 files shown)


Sound/Tidal/Pattern.hs
```diff
@@ -1457,9 +1457,8 @@ to change this use `toScale' size`.  Example:
 -}
 toScale' :: Int -> [Int] -> Pattern Int -> Pattern Int
 toScale' o s = fmap noteInScale
-  where note x = x `mod` length s
```
telephon: Normally, the `octave` is not the length of the scale.

erwald: True, but neither does the `octave` function – the `octave` function gets the octave to which the given input needs to be mapped.

Or do you mean that the argument `o` ought to be renamed?

bgold-cosmos: @telephon the `o` argument is meant to be "how many semitones before the scale repeats".  Or more literally "what offset should be used for note indices outside the given scale", e.g. if the scale is `[0, 4, 7]` what note should be returned for the 4th element.  That's not exactly what an "octave" is, but much of the time it's 12 anyway.

telephon: yes, that's about what I would have expected. But that is not `length s`, is it?

erwald: @telephon Just to clarify, are you referring to the argument `o` or the function `octave`? You're saying that one of those isn't the same as `length s`? Is it a question of nomenclature or correctness of implementation?

telephon: It's fine, sorry - I misread the function implementation (I should have looked at the whole file, not just the changes).

erwald:
@bgold-cosmos Ready to be merged in?

PR #235, opened 2017-09-15
Add two Bjorklund functions

bgold-cosmos:
Adds `einv` which fills the "blanks" left by `e`:

`e 3 8 "x"` -> `"x ~ ~ x ~ ~ x ~"`
`einv 3 8 "x"` -> `"~ x x ~ x x ~ x"`

And `efull n k pa pb` which takes two patterns, and stacks `e n k` on
the first with `einv n k` on the second.

`efull 3 8 "2" "1"` -> `"2 1 1 2 1 1 2 1"`

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -1129,9 +1129,16 @@ e n k p = (flip const) <$> (filterValues (== True) $ listToPat $ bjorklund (n,k)
 e' :: Int -> Int -> Pattern a -> Pattern a
 e' n k p = fastcat $ map (\x -> if x then p else silence) (bjorklund (n,k))
 
+{- | `einv` fills in the blanks left by `e`
+ -
+ @e 3 8 "x"@ -> @"x ~ ~ x ~ ~ x ~"@
+
+ @einv 3 8 "x"@ -> @"~ x x ~ x x ~ x"@
+-}
 einv :: Int -> Int -> Pattern a -> Pattern a
 einv n k p = (flip const) <$> (filterValues (== False) $ listToPat $ bjorklund (n,k)) <*> p
 
+{- | `efull n k pa pb` stacks @e n k pa@ with @einv n k pb@ -}
 efull :: Int -> Int -> Pattern a -> Pattern a -> Pattern a
 efull n k pa pb = stack [ e n k pa, einv n k pb ]
 
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -1129,6 +1129,11 @@ e n k p = (flip const) <$> (filterValues (== True) $ listToPat $ bjorklund (n,k)
 e' :: Int -> Int -> Pattern a -> Pattern a
 e' n k p = fastcat $ map (\x -> if x then p else silence) (bjorklund (n,k))
 
+einv :: Int -> Int -> Pattern a -> Pattern a
+einv n k p = (flip const) <$> (filterValues (== False) $ listToPat $ bjorklund (n,k)) <*> p
+
+efull :: Int -> Int -> Pattern a -> Pattern a -> Pattern a
+efull n k pa pb = stack [ e n k pa, einv n k pb ]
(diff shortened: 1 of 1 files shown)


Sound/Tidal/Pattern.hs
```diff
@@ -1129,6 +1129,11 @@ e n k p = (flip const) <$> (filterValues (== True) $ listToPat $ bjorklund (n,k)
 e' :: Int -> Int -> Pattern a -> Pattern a
 e' n k p = fastcat $ map (\x -> if x then p else silence) (bjorklund (n,k))
 
+einv :: Int -> Int -> Pattern a -> Pattern a
+einv n k p = (flip const) <$> (filterValues (== False) $ listToPat $ bjorklund (n,k)) <*> p
```
telephon: somehow it seems that we might be missing another function that replaces one thing in a pattern with another?

almost like one that just exchanges the gaps  (untested)

```haskell
gapswap :: Pattern a -> Pattern a -> Pattern a
gapswap p q = (flip const) <$> (filterValues (== False) $ p) <*> q
```

Then this would become
```haskell
einv n k p = gapswap  (listToPat $ bjorklund (n,k)) p
```

and there could be an equivalent of `efull`
```haskell
full :: Pattern a -> Pattern a -> Pattern a
full pa pb = stack [ pa, gapswap pa pb ]
```

bgold-cosmos: Patterns don't have "False" entries in their gaps so `gapswap` doesn't work; in general to determine if a Pattern has gaps at all you need to examine an (arbitrarily long!) interval.

It works OK with the Euclid notation because those make a very specific subset of pattern that _is_ list-like and has well-defined gaps.  But I'm not sure if it's generalizable.

bgold-cosmos: That said, I now recall @kindohm asking to do something like this, and I think someone came up with a weird workaround, so I'll try to see if I can find out how that worked.

telephon: I see, yes I wondered already about `(== False)`.  So when you write `s $ "bd ~ hh ~"`, the `~` is basically ignored?  It would make sense to have a designated instrument name for a gap (`~`). The you could see that a pause / rest was intended and treat it as an object.

telephon: >in general to determine if a Pattern has gaps at all you need to examine an (arbitrarily long!) interval.

all you need is to determine whether it has a gap now (or in the current arc), no?

bgold-cosmos: The problem is what happens with `stack ["bd ~", "~ sn"]`?  The individual patterns have gaps, but you'd expect the stack to be gapless.  If "gap" becomes an event, then either pattern-merging functions have to check and potentially remove them, or you have merged patterns with loads of now-meaningless "gap" events.  I think the very general way Patterns are defined makes this unworkable

And you're right about my interval comment, I was thinking about a slightly different issue that isn't relevant here (querying an arc may return events arbitrarily far in the past).

telephon: I see. well, I suppose a `full` function could be stupid and just fill those gaps anyway, even when stacked.

` stack [ fill a, fill b] = fill stack [a, b]`
would still be valid

yaxu:
Very good to have but I'm thinking about a more general solution, see
https://github.com/tidalcycles/Tidal/issues/250

bgold-cosmos:
Cool, that's why I left this one alone...

Sound/Tidal/Pattern.hs
yaxu: I made a `fill` function that takes the gaps from one pattern and uses them to structure events from another: https://github.com/tidalcycles/Tidal/pull/251/commits/f30ef356bbe9d2bc371fa109caf5c78af2c27a64

It's interesting but indeed doesn't understand polyphony, and will treat two adjacent silences as one.

yaxu:
Still pondering structured silence so thought this better be merged in the meantime!

PR #240, opened 2017-10-12
Topic chord scale

yaxu:
chord and scale patterns

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -1,3 +1,4 @@
+
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# OPTIONS_GHC -Wall -fno-warn-orphans -fno-warn-name-shadowing #-}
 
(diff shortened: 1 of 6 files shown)


Sound/Tidal/Pattern.hs
```diff
@@ -327,7 +328,9 @@ density :: Pattern Time -> Pattern a -> Pattern a
 density = fast
 
 _density :: Time -> Pattern a -> Pattern a
-_density r p = withResultTime (/ r) $ withQueryTime (* r) p
+_density r p | r == 0 = silence
```
telephon: could also be:

```
_density r p | r == 0 = silence
                     | r < 0 = rev $ _density (0-r) p
                     | otherwise = withResultTime (/ r) $ withQueryTime (* r) p
```

yaxu: Thanks very much, I think that's what I meant to do!

PR #244, opened 2017-10-24
Add VideoDirt module for Tidal

filippoguida:
https://github.com/filippoguida/VideoDirt


yaxu:
There's quite a lot of duplication here. I think best to put it in its own package for now, and longer term make tidal more configurable for custom OSC.

PR #253, opened 2017-10-29
use explicit or qualified imports, remove some unused imports

JeffreyBenjaminBrown:


bgold-cosmos:
Sound.Tidal.Dirt now also uses the `temporalParam2` and `temporalParam3` functions from Sound.Tidal.Pattern; there might be a few other new things to 1.0-dev that haven't been picked up.

JeffreyBenjaminBrown:
Accidentally nuked the repo that came from.

PR #254, opened 2017-10-29
use explicit or qualified imports, remove some unused imports

JeffreyBenjaminBrown:


JeffreyBenjaminBrown:
I could have sworn this test was passing, but now it seems not to.

I made a separate (3 lines) commit that speeds up the isSubset function in Stream.hs, and may have submitted a pull request for that separately, or maybe it's bundled with this one; I can't tell. It's not passing the Travis build either.

JeffreyBenjaminBrown:
It would be efficient if I could test the Travis build locally before committing changes. Is that possible?

yaxu:
@lennart can you offer insights on Travis please?

yaxu:
I triggered another build and it still fails unfortunately:
  https://travis-ci.org/tidalcycles/Tidal/builds/294339815

Sound/Tidal/Dirt.hs
```diff
@@ -15,13 +15,13 @@ import Data.Fixed
...
-import Sound.Tidal.Pattern
-import Sound.Tidal.Parse
-import Sound.Tidal.Params
-import Sound.Tidal.Time
-import Sound.Tidal.Tempo
+import Sound.Tidal.Stream (Backend(..), Shape(..), Value(..), ParamPattern(..), state, latency, cpsStamp, params, follow, copyParam, (|*|), (#), (|*|), (|+|), setter, stream)
+import Sound.Tidal.OscStream (TimeStamp(..), OscSlang(..), makeConnection, namedParams, path, timestamp, preamble)
+import Sound.Tidal.Pattern (Pattern(..), envL, rotR, temporalParam, fastcat, atom, spread', overlay, envEq, envEqR, silence, _slow, unwrap, stack, fadeIn', fadeOut')
+import Sound.Tidal.Parse (ColourD(..), p)
+import Sound.Tidal.Params (s_p, orbit_p, n_p, gain, gain_p, loop, sound, end, begin, release_p, hold_p, attack_p, loop_p, unit_p, bandq_p, bandf_p, hresonance_p, hcutoff_p, coarse_p, crush_p, delayfeedback_p, delaytime_p, delay_p, cut_p, kriole_p, shape_p, accelerate_p, resonance_p, cutoff_p, vowel_p, velocity_p, pan_p, speed_p, end_p, begin_p, offset_p)
```
vivid-synth: I find this less readable. If we're importing more than, say, 5 items, can we not just import the module unqualified? (Particularly if it's a module in the same package)

JeffreyBenjaminBrown: I don't feel strongly.

If all you want to know is which modules are involved, then you can skip the words in parentheses, and it's no less readable than if the imports weren't qualified. The advantage, though is that if you want to know, you can tell which parts of what modules are getting used in one place, rather than having to go through the whole file and for each function run grep or something to find out where it came from.

Maybe it's only useful if you want to do dependency analysis or plan on subdividing or rearranging modules. I was doing both -- I was studying Tidal, and found rearranging it helpful for that.

But I think I'm done studying it, and it's probably evolving rapidly enough that these changes are stale by now. And I'm not using the rearrangements either -- I didn't want my code to become incompatible with the master branch, so instead I just reimported and in some cases renamed things from Sound.Tidal into Sound.Tidal.Epic.

vivid-synth: Another way to find where a function came from is to do e.g. ```:i hresonance_p``` in ghci/tidal

JeffreyBenjaminBrown: Yeah, that was usually what I did. But I remember grepping too. Maybe I was trying to find all the files where a function was used.

yaxu:
Re discussion here seems there isn't a strong case for this https://github.com/tidalcycles/Tidal/pull/254/files
It would still be good to remove unused imports, but we can do that later as part of a general linting excercise.

Issue #255, opened 2017-10-31
(suggested refactoring) Sound.Tidal.Parse.toPat should not require Parseable constraint 


jwaldmann:
I was wondering why 
```
toPat :: Parseable a => TPat a -> Pattern a
```
when we call this function, the parser is finished, as it did produce the `TPat a`. Of course the parser needed the type constraint, but later processing steps should not. When I remove the constraint from the type,
I get an error for `fromTo` in this branch of `toPat`
```
TPat_EnumFromTo a b -> unwrap $ fromTo <$> (toPat a) <*> (toPat b)
```
This is because of
```
class Parseable a where
  parseTPat :: String -> TPat a
  fromTo :: a -> a -> Pattern a
  fromThenTo :: a -> a -> a -> Pattern a
```
It seems to me that enumeration is not part of parsing, and this class could be split up in two
```
class Parseable a where
  parseTPat :: String -> TPat a
class Enumerable a where
  fromTo :: a -> a -> Pattern a
  fromThenTo :: a -> a -> a -> Pattern a
```

jwaldmann:
... and then `toPat :: Enumerable a => TPat a -> Pattern a`

yaxu:
This seems to have been done a while back

PR #256, opened 2017-10-31
split class Enumerable off class Parseable, see issue #255

jwaldmann:


yaxu:
Hi @jwaldmann sorry about the delay in looking at this, it has been a busy time..
Looking at this it seems there are other changes mixed in, including a new dependency that looks quite heavy. Is this intentional?

yaxu:
Is it possible to remove the additional commits from this PR @jwaldmann ?

PR #260, opened 2017-12-02
Elisp fixes for distribution of tidal.el as a package

purcell:
I'd like to add `tidal.el` to [MELPA](https://melpa.org/) for more convenient installation, so I went through and fixed up the various packaging issues, and a few other little problems along the way.

### DIFF
--- a/tidal.el
+++ b/tidal.el
@@ -58,6 +58,9 @@
   t
   "*Flag to indicate if we are in literate mode (default=t).")
 
+(defvar tidal-modules nil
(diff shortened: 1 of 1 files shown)


yaxu:
Nice, thanks!

yaxu:
(Will merge once I fix the unrelated build error..)

purcell:
Great, I've just added a recipe for `tidal.el` in MELPA, so packages will get auto-built there starting soon.

PR #278, opened 2018-02-02
generalize `toScale` so it works with `up`

bgold-cosmos:
`toScale` should now be able to produce Patterns of any numeric type, not just Pattern Int


telephon:
related: https://github.com/musikinformatik/SuperDirt/pull/72

yaxu:
I'm going to go ahead and accept this as it's needed for `n` as well as `up` now

yaxu:
(thanks @bgold-cosmos!)

PR #293, opened 2018-04-20
add params for dirt's leslie

reprimande:
Hi.

SuperDirt have the effect called 'leslie'.
https://github.com/musikinformatik/SuperDirt/blob/aa6b7d3ae72afc483fbc7e45bcbe5e4a9379a647/synths/core-synths.scd#L274-L315

But Tidal doesn't support that yet.
I want to use that by default.

Please validate it.


yaxu:
Thanks very much for this! Development work currently takes place in the 1.0-dev branch, not master, so I reverted the PR. I'll re-apply the patch there for you.

reprimande:
Thanks for checking this! And sorry without understanding the rule of branch...
I'm looking forward to you'll re-apply the patch!

PR #305, opened 2018-05-22
1.0 dev

yaxu:
Merge dev branch into master, lets do development work there from now on, and use tags for releases.


yaxu:
Actually would be good to resolve current PRs before switching to master.

Issue #313, opened 2018-07-01
"within" not functioning as it should


TylerMclaughlin:
When I type ` d1 $ within (0, 0.2) (fast 2) $ sound "bd hh cp sd"`,
 I expect the equivalent of `d1 $ sound "[bd bd] hh cp sd"` but instead I hear `d1 $  sound "[bd hh] hh cp sd"`.

Via the [reference](https://tidalcycles.org/functions.html#within) for the `within` function:

 ```d1 $ within (0, 0.5) (fast 2) $ sound "bd*2 sn lt mt hh hh hh hh"```

When I copy this code and run it, it is different than the reference audio sample, instead applying `fast 2` to the hi-hats.

bgold-cosmos:
The behavior is what I'd expect, the reference audio sounds wrong to me!

`within (0, 0.2) (fast 2)` to me means "use `fast 2` for the first two tenths of a cycle, then use the original pattern for the rest".  `fast 2 $ s "bd hh cp sd"` is `"bd hh cp sd bd hh cp sd"`, so you get the first two beats of that before reverting to the original.

I don't know why the reference audio sounds weird.

TylerMclaughlin:
Hi @bgold-cosmos!

Thanks to your reply I realize now that there are two ways of thinking about how this could work because there are two ways to chain up `within` and `fast` that sound different.     

The way you're thinking about it (strategy A) is that first the original cycle is cloned (duplicated).  Then `fast` is applied to that cloned cycle yielding `"bd hh cp sd bd hh cp sd"`.  Then the arc corresponding to the  [0, 0.2] interval in the new, faster  cycle is swapped for the [0,0.2] interval arc in the original slower cycle.   This is what tidal is doing currently but not what the reference audio does.

Strategy A in pseudo code:
`Clone cycle -> Apply function to cloned cycle  -> Select interval in cloned cycle -> swap intervals`

The way I was thinking about it (strategy B) is that first the arc of notes in just the [0,0.2] interval in the original cycle are selected, producing (conceptually) a masked version `"bd ~ ~ ~"` .  Then `fast` is applied to the the non-rest part, producing `[bd bd] ~ ~ ~`.    Then that part is swapped with the original notes in the [0,0.2] interval.

Strategy B in pseudo code:
`Select interval in cycle -> clone interval -> apply function to cloned interval -> swap intervals`


New to Haskell AND new to tidal btw so please forgive me if I'm making a fool out of myself.

[...]

PR #316, opened 2018-07-28
adding within' function

TylerMclaughlin:
Issue #313. 

Incorporating code from @bgold-cosmos following our discussion.   Added comment in code for explanation of what this function does.

### DIFF
--- a/Sound/Tidal/Pattern.hs
+++ b/Sound/Tidal/Pattern.hs
@@ -1116,6 +1116,44 @@ within (s,e) f p = stack [playWhen (\t -> cyclePos t >= s && cyclePos t < e) $ f
                           playWhen (\t -> not $ cyclePos t >= s && cyclePos t < e) $ p
                          ]
 
+{- |
+For many cases, @within'@ will function exactly as within.  
+The difference between the two occurs when applying functions like 'fast' or '<~'. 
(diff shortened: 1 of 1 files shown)


Sound/Tidal/Pattern.hs
```diff
@@ -1116,6 +1116,44 @@ within (s,e) f p = stack [playWhen (\t -> cyclePos t >= s && cyclePos t < e) $ f
                           playWhen (\t -> not $ cyclePos t >= s && cyclePos t < e) $ p
                          ]
 
+{- |
+For many cases, @within'@ will function exactly as within.  
+The difference between the two occurs when applying functions like 'fast' or '<~'. 
+within applies the function to the whole cycle first then keeps notes only in the specified interval. 
+within', however, only sees the notes in the specified interval to which it then applies the function.
```
telephon: I'd write:

within' keeps notes only in the specified interval first, and then applies the function.

This makes the symmetry clear.

TylerMclaughlin:
@yaxu @telephon what do y'all think?

yaxu:
I'm traveling back from holidays but am in favour! Longer term it would be
good to generalise this sort of thing though, when we understand it better.

yaxu:
Merged - thanks!

PR #318, opened 2018-07-31
arpg, a function to arpeggiate

pd3v:


yaxu:
Nice idea! I'm definitely in favour of this, but are there other kinds of arpeggios we could think about + support? I don't know much about synth arpeggiators or the theory behind them, but have seen options to go up, down or in random order.. and sometimes arpeggiators seem to be doing something more complex..

pd3v:
Hi @yaxu, I think that the present tidal's reference is already good to create interesting variations on a given arpeggio. For instance, functions like `iter`, `shuffle`, `scramble`, `within`. From here new ideas might flourish.
[An example.](https://github.com/pd3v/soundclips/blob/master/39.tidal)

PR #328, opened 2018-09-12
Get testsuite running!

vivid-synth:
You can run this either with

    cabal test --show-details=streaming

or

    ghci -isrc test/Test.hs # and then call 'main'

There are a couple tests you might want to rename (e.g. ```it "one" $``` haha)


yaxu:
Thanks!

PR #331, opened 2018-09-12
more on arpeggios' functions 

pd3v:
Hey @yaxu, some more arpeggiating functions taking into account your suggestions. Please check their relevancy and their naming. I'm also finishing a text (will be a gist) with a short explanation of what arpeggios are, then how to tidal code with these functions and bring out arpeggios potencial and if possible bend them.

### DIFF
--- a/Sound/Tidal/Chords.hs
+++ b/Sound/Tidal/Chords.hs
@@ -161,6 +161,32 @@ chord :: Num a => Pattern String -> Pattern a
 chord p = flatpat $ chordL p
 
 -- | @arpg p@ turns a pattern of chord names into a pattern of arpeggios
--- of the those chords respectively
+-- of those chords respectively
 arpg :: Num a => Pattern String -> Pattern a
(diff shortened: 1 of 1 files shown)


pd3v:
This is the gist with the text mentioned above about arpeggios coding in tidal -> [](https://gist.github.com/pd3v/758ddab0e5c8017f2bce9454dd6ab573)

pd3v:
Thanks! Ok, I'll be around.

yaxu:
Hi it looks like this isn't compiling
Due to a missing `import Sound.Tidal.Stream` but also an undefined function `n` used in a few of the functions
What should `n` do?

yaxu:
ah I see, it's the parameter. I've done a quick fix for now, we should pay more attention to travis!

pd3v:
Aah, precious help from you! Sorry for those imports' extra work. I've checked (now) travis job log. Thank you.

PR #366, opened 2018-10-31
Patternify

msp:
WIP - Do not merge!

Opening this early to get feedback as we go :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,10 +47,15 @@ add tests for any new functionality. You can look for things that need
 testing like this:
 
 ```
+cabal install --only-dependencies
 cabal configure --enable-coverage    # only need to do this the first time
 cabal test --show-details=streaming
 firefox dist/hpc/prof/html/tests/hpc_index.html
 ```
+
+To run up your changes locally, install Tidal with `cabal install`. To remove
+them again and revert to the latest release, run `ghc-pkg unregister tidal-1.0.0`
+being sure to match up the version numbers.
 
 # A process for making a release
 
--- a/src/Sound/Tidal/MiniTidal.hs
+++ b/src/Sound/Tidal/MiniTidal.hs
@@ -241,7 +241,7 @@ patternTransformationWithArgs = parensOrNot $ choice [
   (function "degradeBy" >> return T.degradeBy) <*> patternArg,
   (function "unDegradeBy" >> return T.unDegradeBy) <*> patternArg,
   (function "degradeOverBy" >> return T.degradeOverBy) <*> literalArg <*> patternArg,
-  (function "sometimesBy" >> return T.sometimesBy) <*> literalArg <*> patternTransformationArg,
+  (function "sometimesBy" >> return T.sometimesBy) <*> patternArg <*> patternTransformationArg,
   (function "sometimes" >> return T.sometimes) <*> patternTransformationArg,
   (function "often" >> return T.often) <*> patternTransformationArg,
   (function "rarely" >> return T.rarely) <*> patternTransformationArg,
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -163,8 +163,8 @@ almostNever = sometimesBy 0.1
 almostAlways = sometimesBy 0.9
 @
 -}
-sometimesBy :: Double -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
-sometimesBy x f p = overlay (_degradeBy x p) (f $ _unDegradeBy x p)
+sometimesBy :: Pattern Double -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
+sometimesBy x f p = overlay (degradeBy x p) (f $ unDegradeBy x p)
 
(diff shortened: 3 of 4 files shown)


msp:
On that note.. any feedback on the tests? 

I prefer the `compareP` style.

Is there a clearer way to create patterns other than the combination of `fast cat` and `pure` ? Thought I'd seen something like tagging or 'casting' a string somewhere..

bgold-cosmos:
Interactively, if Tidal didn't magically turn note names into MIDI numbers, `fastcat ["a", "b", "c"]` would work just fine, the `pure` isn't necessary because of string overloading.  If you're using ghci  with string overloading, you can avoid the MIDI magic with explicit typing, which is tedious:
`fastcat ["a"::Pattern String, "b"::Pattern String]`

If you're not using string overloading, you can invoke the parser explicitly - but if you still want to avoid the automatic typing of literals you need to do something like
```
pS = toPat . (parseRhythm pVocable)
```
and then
```
fastcat $ map pS ["a", "b", "c"]
```
should work.

yaxu:
I'm really inexperienced with tests myself, and was just taking the approach of writing now and tidying up later.  Feel free to play around or make changes.
The main reason why there is so much `fastcat`, `pure` etc is because I wrote the existing tests before porting over the parser! Feel free to use `"a b"` instead of `fastcat [pure "a", pure "b"]`

As @bgold-cosmos says, you might have to give hints about types. Haskell is really good at inferring types though, so instead of `fastcat ["a"::Pattern String, "b"::Pattern String]` you only have to do give it enough of a hint to work out the rest, e.g. `fastcat ["a"::Pattern String, "b"`, or `fastcat ["a", "b"] :: Pattern String`

yaxu:
By the way, I don't think it's the case that note parsing is messing with the types. By the time a string is being automatically parsed into a pattern, Haskell has already decided what its type is. If it doesn't know the type, I think it defaults to an integer, therefore invoking the note parsing. The contents of the string isn't a factor in the type of the resulting pattern.

yaxu:
One more thing, you _could_ do tests for `sometimesBy` etc.. As random patterns are deterministic you'll always get the same values out for same timespans.

yaxu:
I meant tests for `sometimesBy 0.5` etc

msp:
Hmm, I can't get any of the compiler hint variants to work in the tests @yaxu - they all end up seeing either "a", "b" or "c" as a Char. 

I also looked at using the `Pattern` constructor but that'd prob end up more complicated by the time we've wired up the dependent types.

Stepping back a bit, my motivation is to have self documenting tests that the end user can review so they should resemble the DSL. We could have public/doc tests plus internal tests where readability is not a concern. 

Thoughts on something like https://github.com/tidalcycles/Tidal/pull/366/commits/6cf31754d845f87c0866f27f0792a78ca6e6a244  which uses the `Parser` directly as @bgold-cosmos suggests?

yaxu:
@msp I've added an example that implicitly uses the parser at the end.

Your `pattern` function there is simply `p` from `Sound.Tidal.Parse` but with a constrained type.. So you could define it like this:
```
pattern :: String -> Pattern String
pattern = p
```
Maybe `ps` is a better name.

msp:
Played around with a test for `sometimesBy 0.5` and found a couple of odd things. Here's the test:

```
    it "applies the 'rev' function sometimes when set at 50% probability" $ do
      let
        overTimeSpan = (0, 3)
        testMe = sometimesBy 0.5 (rev) (ps "bd bd hh sn")
        expectedResult = cat ["bd bd hh sd",  "sn hh bd bd", "bd bd hh sn"]
        in
          compareP overTimeSpan testMe expectedResult
```

Results in actual vs expected:

```
[(0>¼)|"sn",(¼>½)|"bd",(½>¾)|"hh",(¾>1)|"bd",(1>1¼)|"sn",(1¼>1½)|"hh",(1½>1¾)|"bd",(1¾>2)|"bd",(2>2¼)|"bd",(2½>2¾)|"bd",(2½>2¾)|"hh",(2¾>3)|"sn"]
vs
[(0>¼)|"bd",(¼>½)|"bd",(½>¾)|"hh",(¾>1)|"sd",(1>1¼)|"sn",(1¼>1½)|"hh",(1½>1¾)|"bd",(1¾>2)|"bd",(2>2¼)|"bd",(2¼>2½)|"bd",(2½>2¾)|"hh",(2¾>3)|"sn"]
```

1) What's with the order jumbling in the first cycle? I'd expect `bd bd hh sn` or the reverse of that?
2) The Arcs for 3rd cycle seem out:  `"bd",(2>2¼)|"bd",(2½>2¾)` - it misses `(2¼> 2½)` ? 

Did I mess up the test?

yaxu:
Hmm, I think such strange happenings when you combine `sometimesBy` with `rev` are expected. The sometimes applies to every event separately. So one event might be reversed, but the event in the space where it ends up might not be reversed.

yaxu:
I do like the idea of making the tests as readable as possible

yaxu:
Thanks a lot @msp!

msp:
You're welcome @yaxu ! It's slow going but it's forcing me to get my head into the types and their various contexts so it's a good learning experience.

PR #367, opened 2018-11-01
Split out tests to individual files

msp:


msp:
Yea, exactly. Makes it easier if we want to use the UI function tests as extra docs also.

What's your preferred system? Do you merge in PRs or do you prefer contributors to do it?

yaxu:
We don't really have a system. What do you think is best? I suppose once a PR is reviewed it doesn't matter too much who presses the button

msp:
Yea exactly, so long as only one approval is needed which I'm guessing is the case here. I'll do it now but shout if you prefer otherwise in future.

PR #372, opened 2018-11-04
Improve error handling in the BP2 style parser

### LINKED ISSUE #371: Deal with sequence parser failure
The polyrhythmic sequence parser `p` currently returns `silence` on syntax error, e.g. missing bracket. It would be much better to raise an error in some way, as silence is a really bad failure mode and there is no error reporting to the end-user programmer. Instead it would be better to either raise an exception for people in IO land (e.g. ghci), or return something like `Either String TPat` for people wanting a purer parse (e.g. @d0kt0r0 in minitidal).

I also think we should reconsider #175 and rename `p` to something else, as it's almost never called directly and so `p` is a really useful identifier that's never used. Furthermore according to #370, there may be more than one little sequence language in the future, so a more descriptive name for this one would be good. It's highly inspired by the Bernard Bel's BP2 syntax, so howabout..

```
parseBP_AST :: Parseable a => String -> Either String (TPat a)
parseBP :: Parseable a => String -> Either String (Pattern a)
parseBP_E :: Parseable a => String -> Pattern a
```
... where the latter returns the right value, or turns the left value into an exception.

Related to #131.

yaxu:
To fix #371


yaxu:
Note that the raised exceptions are raised but aren't caught yet - currently causing the tidal scheduler to crash out!

PR #387, opened 2018-11-14
Generalise `rand` to match other continuous functions

### LINKED ISSUE #386: Generalise 'rand' to match saw, sine et al
Currently `rand` is `Pattern Double` which means it can't be used for time patterns etc. Should be generalised to `rand :: Fractional a => Pattern a`, so it can always be swapped in from `sine` et al.

msp:
Fixes #386


yaxu:
Thanks! still travelling but will try to have a look later

PR #389, opened 2018-11-15
More Atom-friendly BootTidal.hs (sets OverloadedString option, less p…

bgold-cosmos:
…rompt

spam)


bgold-cosmos:
Sorry, ignore the Control.hs changes, that was supposed to be a separate PR

PR #390, opened 2018-11-16
default option for `cI` too

bgold-cosmos:
To match the behavior of `cF` and others. Also added a `cI1` function (a common default would be one for integer patterns, such as for `striate`, `stut`, etc.)


yaxu:
Thanks! I'm not sure how useful cF0 etc will turn out to be.. cF 0 is only one key press away..

PR #391, opened 2018-11-16
Do /127 conversion in cF for incoming Int messages

bgold-cosmos:
If the incoming OSC value is integer format, and Tidal is converting it to a Pattern Double, then it's likely a MIDI value >float conversion.  So dividing by 127 here is more convenient for the end user - with this change both `cI` and `cF` will produce reasonable results from incoming Integer messages.

### DIFF
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -357,7 +357,7 @@ _cX f ds s = Pattern Analog $
 
 _cF :: [Double] -> String -> Pattern Double
 _cF = _cX f
-  where f a (VI v) = [((a,a),fromIntegral v)]
+  where f a (VI v) = [((a,a),(fromIntegral v)/127.0)]
         f a (VF v) = [((a,a),v)]
         f a (VS v) = maybe [] (\v' -> [((a,a),v')]) (readMaybe v)
 


yaxu:
I somehow merged this in error. This patch assumes that incoming osc comes from midi which isn't necessarily the idea. I think any midi scaling should go in midi specific code (which currently lives in the pd patch). Ok if I back this out, @bgold?

bgold-cosmos:
It's OK, but it only does the scaling for the specific case where the OSC message value has integer type but `cF` is being used to convert to float.  Even if the original source wasn't MIDI I can't imagine a case where a Tidal user would want to convert an incoming integer to float without scaling, but I'm sure there are use cases I haven't thought of.

To contrast, the pd patch sends OSC floats, and doesn't really work with `cI` as is, since it applies scaling no matter what (and anything in the 0-1 range just gets `floor`ed to zero)

bgold-cosmos:
Actually, now that I think about it, part of the "problem" I was trying to solve is that I would really like context-sensitive scaling, which is impossible to put in MIDI specific code since the sender/translation layer has no idea what the destination is.  Only Tidal can know, so I had to put the code in `cF`.

But that doesn't mean it's a good idea - it might be worth thinking about what use cases would look like.

yaxu:
The case I have in mind is languages with duck typing, where the type isn't important. I think tidal should do its best to coerce osc types cleanly whenever possible. In particular I think this is in the spirit of how supercollider handles OSC.

If you really want this scaling to happen in the OSC receiver rather than sender, perhaps a cleaner approach would be adding `cCC` or similar, that does this scaling. But really I see this 127 as an arbitrary (and low) granularity value which MIDI doesn't always respect anyway afaik.

The pd patch is just meant as a quick example, I don't understand how types work there and just hacked up something to things working. We can make a more direct MIDI interface later, this is an API useful for interacting with any language that supports OSC.

Issue #345, opened 2018-09-24
Conditional Modulations


cannc4:
I find this function written by @bgold-cosmos to be super useful and it might be a good idea to take a look at its inner workings. It's sort of an hacky `conditional` function

```
modNote n f p = stack [f $ filterValues (isEq n . (Map.!?n_p)) p,filterValues (not . isEq n . (Map.!?n_p)) p] where isEq a m = (isJust m) && (a == (fromJust $ fromV $ fromJust m))

modValues par test f p = stack [ f $ filterValues (test' . (Map.!?par)) p, filterValues (not . test' . (Map.!?par)) p ] where test' m = (isJust m) && (test (fvalue $ fromJust m))
```

yaxu:
I think this _might_ be a problem for Haskell's lenses to solve, but I'm not sure as I don't understand lenses just yet..

Perhaps for now we should just include these functions as-is. I propose renaming them to avoid confusion with modulo. Perhaps `withNote` and `withControl`?

yaxu:
We have two approaches to this:
https://github.com/tidalcycles/Tidal/pull/392

- `withControlValue`, `withControlValues`, and `withControlRange`: https://github.com/tidalcycles/Tidal/pull/392
- `fix`, `unfix` and `contrast`: https://github.com/tidalcycles/Tidal/commit/55ee834048386450d3fcf5e71143dc3cefb089db

I think `fix` is more flexible and easier to use than `withControlValue`.
```
d1 $ withControlValue "n" 3 (# crush 3) $ s "supermandolin" >| n "1 2 3"
d1 $ fix (n 3) (# crush 3) $ s "supermandolin" >| n "1 2 3"
```
It might be best to only go with `fix` et al for simplicity and patternability, but it probably can't support all the features of `withControlValue` et al.. Thoughts?

bgold-cosmos:
(continuing discussion from #392)

TolerantEq is in Pattern.hs, line 214 I think.  I think your implementation of fix is structurally the same as what I was calling `whereCV`, for reference:
```
maskr pb p = filterJust $ filt <$> pb Sound.Tidal.Context.*> p where
  filt True b = Just b
  filt False b = Nothing

whereCV m f p = stack [
  f $ maskr pb p,
  maskr (fmap not pb) p
  ] where
    pb = liftA2 (~==) m p
```

[...]

PR #392, opened 2018-11-16
Value type

bgold-cosmos:
Again, sorry for the reverts in the history, trying to do too many PRs at once...

This implements a `filterControlValues` function that works much like `filterValues` but operates on ControlPatterns and makes it a bit easier to pick out specific values.  For this to work on generic types it is unfortunately necessary to resurrect the ParamType class (now called ValueType), so there's a little extra bolierplate in Pattern.hs

There are also `withControlValue`, `withControlValues`, and `withControlRange` functions, which generalize the functionality of `modNote` mentioned in #345

### DIFF
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -357,7 +357,7 @@ _cX f ds s = Pattern Analog $
 
 _cF :: [Double] -> String -> Pattern Double
 _cF = _cX f
-  where f a (VI v) = [((a,a),(fromIntegral v)/127.0)]
+  where f a (VI v) = [((a,a),fromIntegral v)]
         f a (VF v) = [((a,a),v)]
         f a (VS v) = maybe [] (\v' -> [((a,a),v')]) (readMaybe v)
 
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -357,7 +357,7 @@ _cX f ds s = Pattern Analog $
 
 _cF :: [Double] -> String -> Pattern Double
 _cF = _cX f
-  where f a (VI v) = [((a,a),fromIntegral v)]
+  where f a (VI v) = [((a,a),(fromIntegral v)/127.0)]
         f a (VF v) = [((a,a),v)]
         f a (VS v) = maybe [] (\v' -> [((a,a),v')]) (readMaybe v)
 
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -345,21 +345,15 @@ _stut' :: (Num n, Ord n) => n -> Time -> (Pattern a -> Pattern a) -> Pattern a -
 _stut' count steptime f p | count <= 0 = p
                           | otherwise = overlay (f (steptime `rotR` _stut' (count-1) steptime f p)) p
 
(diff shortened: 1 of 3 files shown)


yaxu:
Sounds good, I'm going to commit unread so I can have a look while offline todayx

yaxu:
Reading this on the train, I'm not sure what problem this is solving.. do you have a use case or two?

bgold-cosmos:
It's useful for applying effects only to specific notes, or similar:
```
d2
 $ withControlValues "n" [-19,-20] (# shape 0.8)
 $ s "superzow"
 >| n ("[g5(<3 5 3 3>,8), b4(5,8,<2 2 3 0>), es3(7,9,3)]" + "<0 -1>/16")
 # cutoff 800
```

There are ways to do this with only vanilla `filterValues`, but AFAICT they get pretty awkward and tricky to set up quickly during a performance.

yaxu:
Thanks, understood. (I missed the first parameter and thought this was working on _all_ values)

yaxu:
Ok looking at this.. safeTest doesn't look too safe.. In your example there's nothing to say that `[-19,-20]` are doubles or ints, so I'm surprised if `fromV` doesn't crash out..
See the second test here for the core of the problem: https://github.com/tidalcycles/Tidal/commit/4e6eb7fa7fb8c065c878b341a18d881bf671654e

pd3v:
> Ok rejigged it to be safer, what do you think? I think it would be nice to find shorter names for these functions, any ideas?
> [2469ee1](https://github.com/tidalcycles/Tidal/commit/2469ee17958cbc340cff63827f1f5b4f3568cc7b)

@bgold-cosmos, great ideas. 

A proposal for shorter names:
filterControlValues -> fltrCtrlVal or fltrCtrlV
withControlValues -> withCtrlVal or withCtrlV
withControlRange -> withCtrlRng or withCtrlR

bgold-cosmos:
@yaxu OK, I'll check that it works but that looks like it should be good.  I think only one out of `toV`/`fromV` is ever necessary, and you're right that `toV` is safer.

About abbreviations - would CV be an understandable shorthand for ControlValue?  It has the advantage (disadvantage?) of already being a term of use in audio.  So then we could have filterCV, withCV, withCVRange, etc.

yaxu:
Thanks @bgold-cosmos, I'll remove `fromV` for now then
We have both `Parseable` and `Valuable` classes now, I'll make an issue to consider merging them, although I'm not sure if that's possible/desirable.
I think it's good to prioritise memorability in  naming things, and think `filterCV` is probably easier to remember than `fltrCtrlVal` for example. I have an idea for a different approach though (something like `where (s "bd") (# crush 3) $ s "bd sn"`) so hang fire on this

bgold-cosmos:
Ah, pattern-based matching & replacement?  I think since Events are already TolerantEq then something like `liftA2 (~==) (s "a b") (s "a b b" # n "1 2")` is the beginning of it.  Can't seem to figure out how to use that for appropriate masking, though.  There are weird structure issues.

Actually, I think `mask` may be buggy too.  Looking into it...

yaxu:
`TolerantEq` sounds interesting, I can't find it though
Maybe let's continue discussion on #345

yaxu:
@pd3v see #345, we're looking at shorter still `fix` for this

PR #398, opened 2018-11-23
Updating arpeggiating functions for tidal 1.0.0

pd3v:


yaxu:
Thanks for this,
I like `arpgt'`, but I think using `rot` here isn't quite right, as that rotates a whole cycle and not notes within a chord. I've been thinking of a way to do this.

The others I'm less sure about, `arpgt "c'maj"` for example is shorthand for `n (arpg "c'maj")` but there is also `note (arpg "c'maj")`.. I think it might just be less confusing not to hide control names behind shorthands.

`arpgtt rev p` seems to be the same as `n (arpg rev p)`, again I think the latter is more explicit and flexible. The former is shorter but I don't think I would be able to remember what `arpgtt` did.

Sorry to be a downer but I think with this it would be good to work towards a small set of functions that can be flexibly used together in different combinations, rather than 'baking in' functionality into more rigid 'compound' functions.

pd3v:
You’re welcome, it’s been fun.. I think you’re right about compounding, but it was already present on previous version of these functions, less obvious, I guess, due to `breakUp $ chord` abstraction as `arpg` and from 1.0.0 on no longer needed.

Yeah, `arpgt’` is doing what you say. I was using `<<~` in previous version. Sorry, I didn’t check `rot` properly. 

The idea behind `arpgt` naming is that arpgt is the action concealed in `n (arpg p)`. arpg alone seems to me the chord broken into notes. 

And `arpgti` is a function where one can change parts of the chords/pattern, be it rhythmically, pitch or both. But again, is kind of an abstraction of `within` in an arpeggiating context.

Let me do the honors.. closing PR :)

yaxu:
Thanks @pd3v,
Of course we can have compound functions for doing a specific task, there's a lot of possibilities here for wacked-out arpeggiation that you're exploring. I think these ideas just need a bit more development to make them generally useful as standalone functions. 

What does the `t` stand for in `arpgt`? Or is it just a different way of abbreviating `arpeggiate`?

I think `<<~` will have the same problem as `rot`. How to transform a broken chord as if it were a cycle is a tricky problem but I think we can solve it.

pd3v:
Yeah, `arpgt` = arpeggiate and `arpg` = arpeggio.

PR #403, opened 2018-11-30
Fix setcps via streamOnce

bgold-cosmos:
This adds code to `streamOnce` to properly set the cps when using the `setcps` function in BootTidal.hs.  Also factored out the `toMessage` and `doCps` functions which were common to both `onTick` and `streamOnce`.


yaxu:
Brilliant, thanks!

PR #411, opened 2018-12-04
implement nudge

bgold-cosmos:
`nudge` now works as a ControlPattern and there is a "global" `setNudge` function to adjust overall timing of the stream.  They can be used as follows:

```
d1 $ s "bd sn:2" # cps 1 # nudge "0 0.1"

setNudge tidal 0.05
```


yaxu:
Looks perfect!
Just wondering about the name..
How about renaming `setNudge` to `streamNudgeAll`
Then having this in BootTidal.hs
`let nudgeAll = streamNudgeAll tidal`

yaxu:
Merging for now as this is a definite improvement as-is!

yaxu:
Trying out the rename now

PR #415, opened 2018-12-06
(WIP) Refactor to Arc and Event datatypes.

tonyday567:
@yaxu, rather than talk to what I thought would be a useful refactor, I had a go at introducing Arc and Event datatypes rather than type synonyms.  Naming the bits of an Event helped me a lot in understanding what an Event is, and felt better than using fst & snd where you can quickly lose your place in the code.

The type synonym `Part` refactored out pretty easily, and I think this is a good thing - the common usage was to call the two parts of Part, part and whole (just typing that out confused me).  And usage was almost always in the context of an Event, so not much was lost by integrating into the Event datatype.

There's an increase in some boiler-plate: `Arc 0 1` instead of `(0,1)`, and `Event (Arc 1 2) (Arc 3 4) 5` instead of `(((1,2),(1,3)),5)` but clarity is way up.

I also gathered the functions that operate on Arc's and Event's (in Sound.Tidal.Pattern) together - it's more a matter of taste and I can put them back if that's preferable.  But it helped my understanding of the different computational levels involved as you work up to Patterns.

I also snaffled `end` as the name for Arc's second element and renamed the other end to `fin`.  Can reverse that easily.

I didn't take advantage of the refactor to simplify and such, as yet, and thought it best to check whether you are open to this PR.

The obvious next step is to make Arc an applicative.  I think you'd go `data ArcF a = Arc { start :: a, end :: a }` define the functor and then `type Arc = ArcF Time` to recover `Arc` and hide the polymorphism.

`data EventF a b = Event { whole:: a, part :: a, event (or value?):: b }` and `type Event = EventF (ArcF Time)` is a bifunctor and, I suspect, a very useful one.  I know bifunctor may sound fancy and esoteric but it's very easy and practical here.

### DIFF
--- a/test/Sound/Tidal/PatternTest.hs
+++ b/test/Sound/Tidal/PatternTest.hs
@@ -2,18 +2,18 @@
 
 module Sound.Tidal.PatternTest where
 
-import TestUtils
-import Test.Microspec
+import           Test.Microspec
+import           TestUtils
 
-import Prelude hiding ((<*), (*>))
+import           Prelude             hiding ((*>), (<*))
 
-import Data.Ratio
+import           Data.Ratio
 
-import Sound.Tidal.Core
-import Sound.Tidal.Pattern
-import Sound.Tidal.Control
+import           Sound.Tidal.Control
+import           Sound.Tidal.Core
+import           Sound.Tidal.Pattern
 
-import qualified Data.Map.Strict as Map
+import qualified Data.Map.Strict     as Map
 
 run :: Microspec ()
 run =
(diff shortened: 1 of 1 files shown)


bgold-cosmos:
Just a minor comment - I think it'd be much better to have what you called "fin" remain as "end", as that's a user-visible parameter, and instead define Arc as `data Arc = Arc { start :: Time, fin :: Time}` (or just use finish).  Makes a little more sense linguistically too.

yaxu:
Sorry I'm not a very good git driver. I wasn't sure what to do to fix the conflicting file in test/, and github prompted me to do this:
```
git checkout -b tonyday567-arc-hack master
git pull https://github.com/tonyday567/Tidal.git arc-hack
git checkout master
git merge --no-ff tonyday567-arc-hack
```
But that doesn't seem to be the right thing to do, anyway it's all merged now, so I guess we continue this WIP on the master branch !

My initial thoughts just while editing the tests is that it does seem a readability win. My only concern is whether this comes with a performance penalty, but according to my unscientific experiment things seems OK.

yaxu:
Hm, the way I've merged this has messed up authorship and so on. What's the best way forward - I back out the changes and let you resubmit the PR with the fixed `PatternTest.hs`, @tonyday567?

yaxu:
Hm, it looks complicated to unravel this now, and perhaps it's ok as-is, as I can at least see @tonyday567's commits in the history.

One thing - I agree it was a good idea to leave `begin` intact. Not a strong opinion, but I'd suggest `start` and `stop` rather than `start` and `fin`.

PR #418, opened 2018-12-07
include stack in travis.yml + cleanup of cabal steps

tonyday567:
Also add ghc-8.6.2.


tonyday567:
Not sure why the stack part is failing, but I got a good build on an earlier version, so digging into it.

tonyday567:
I think the problem was similar to https://stackoverflow.com/questions/50709844/is-there-an-issue-with-haskell-ghc-8-4-3-on-travis.  `language: haskell` - what can go wrong?!

yaxu:
Thanks!

PR #422, opened 2018-12-09
Arc and Event instances

tonyday567:
Added Functor instance for Arc, and Bifunctor for Event.  Also made a few changes that result from the refactor:

whole replaces eventWhole
part replaces eventPart
event replaces eventValue
fmap replaces mapArc
wholeOnset replaces eventWholeOnset


yaxu:
This looks great, just wondering whether `value` would be better than `event`?

tonyday567:
I'm not sure what happens if I add to this now, with the merge thing above.  I'll need the recent travis fix as well, so I might just push on and we can always burn this PR to the ground when it's all fixed.

So, on my todo list is:

- `value` replacing `event`
- `stop` replacing `finish`
- fix the shadowing issues (whole, part etc is often used as local names in pattern matching causing lots of warnings).  I think (Event w p v) is just fine but let me know if you want more verbose.
- get to `-Wall`

yaxu:
All good, I'm fine with w p v

tonyday567:
Done with the cleanup.

I went ahead and cleaned up some warnings unrelated to this refactor, because it was getting very close to being able to pass -Werror.

yaxu:
Thanks a lot!

PR #424, opened 2018-12-10
Simplify travis cabal testing

### LINKED ISSUE #423: False positive on cabal tests
I just pushed code that broke a test, the new stack build failed properly, but the cabal ones passed:
  https://travis-ci.org/tidalcycles/Tidal/builds/465606118

do you mind having a look @tonyday567? This is probably related to your stack-related tidying

tonyday567:
Fixes #423.

I tried to simplify, and move to the bleeding edge world of `cabal new-*`, with some but limited success.

There is a case in there, with ghc-8.6.3, where `cabal build` works but `cabal new-build` doesn't, and complains about cairo.  It could be something, so I left the old cabal build in, and made the cabal new-build an allowable fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,12 @@ before_cache:
 
 matrix:
   include:
-    - env: BUILD=cabal GHCVER=8.6.2
-      compiler: "ghc-8.6.2"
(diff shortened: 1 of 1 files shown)


yaxu:
Thanks!

Issue #427, opened 2018-12-14
Pattern and Event structure


tonyday567:
So Pattern is pretty much a list of Events, once you strip away State and Nature, and an Event is a value, a whole time arc and a partial arc.

That's a lot of whole time arcs for one pattern.  Is a better representation something like this?

```
data Event a = Event { part :: Arc, value :: a }
data Chunk a = Chunk { whole :: Arc, events :: [Event a] }
data Pattern a = Pattern { nature :: Nature, query :: State -> Chunk a }
```

In other words, does any particular [Event] list created from a Pattern inevitably have a single sensible whole Arc?  Is this whole arc the same as State eg `query (State theArc Empty) === Chunk theArc events`?

Another open question I have is whether any particular Event list created from a Pattern would inevitably be sorted on usage.  If this is so, you can almost get another representation of the Chunk of a Pattern as:

```
data Chunk a = Chunk { cuts :: [Time], values :: Map Int a }
```

where the cuts are the boundaries of the partial arcs, and in sorted order, so that the whole arc of the chunk is Arc (head cuts) (last cuts).

So "1 ~ 2" is then `Chunk [0, 1/3, 2/3, 1] (FromList $ [(0,1), (2,2)])`.

This lines up with the commentary in #406: The (+) operator (for the values) would be `Map.unionWith (+)`.

yaxu:
I think each event needs its own part and whole. Here's an example of different events having different parts/wholes:
```
queryArc (slow "1 2" "a b" :: Pattern String) (Arc 0 4)
[(0>½)|"a",
 0-(½>1)|"a",
 (1>1½)|"a",
 1-(1½>2)|"b",
 (2>2½)|"a",
 2-(2½>3)|"a",
 (3>3½)|"a",
 3-(3½>4)|"b"
```
The first half of the pattern is a window on `"a b"`, which is always 'a'.  The second half is a window on `slow 2 "a b"`, which alternates between the second half of `a` and the second half of `b`.

(I think this behaviour was different in 0.9.x, hm!)

tonyday567:
Ok, cool, there's a lot more going on than I can wrap my head around.  Having show be `queryArc x (Arc 0 1)` led me to oversimplify.

tonyday567:
```
let p1 = slow "1 2" "0 1" :: Pattern Double
let p2 = "0 1" :: Pattern Double

queryArc p1 (Arc 0 2)
[...]

PR #429, opened 2018-12-16
(WIP) Trying to make a Biapplicative instance of Pattern

tonyday567:
continuing from #427

I made a Biapplicative instance of Event and used this to make an Applicative instance.  I pretty much followed https://www.well-typed.com/blog/2018/09/compositional-zooming/ which is the only example of biapplicative out there I could find.  What zooming has to do with patterns is a mystery right now.

To get a monoid thing from an Arc, I had to add a concept of an EmptyArc, so that intersection (based on subArc) could be `Arc -> Arc -> Arc`.  I also needed a unit - something that when intersected with an arc gives you back the arc.  It would be great to use `Arc 0 infinity` but Rational hasn't got an infinity.

The end game is that `<*>` for Pattern could be something like:

```
(<*>) pf px = Pattern Digital $ \st ->
    filter (not . nonEvent) ((<*>) <$> query pf st <*> query px st)
```
which is at least a little bit awesome. Use the spaceship for lists to apply the spaceship for events (intersect the arcs and apply the values), then filter out the non-intersecting events.

`*>` and `<*` are very interesting.  Since these are already used in Control.Applicative, and are famous, I changed them to `|*>` and `<*|`.    `*>` is action then throw away the left computation and just give me the right one.  `|*>` is throw away the left structure and go with the right one, but `<*>` the values.  If Pattern was a biapplicative, then structure would mean the first of the applicatives that operate on the arc elements.  
In https://hackage.haskell.org/package/bifunctors-5.5.3/docs/Data-Biapplicative.html it might be called `|<*>>` meaning throw away the first applicative of the first computation but keep the rest. Maybe you can't throw it away and it's better expressed as compute it, but then stick the old first left applicative in. Regardless none of this exists and I cant find any reference to anyone having thought about it.

Very much a work-in-progress.

Edit: added backtics

### DIFF
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -249,8 +249,8 @@ en ns p = stack $ map (\(i, (k, n)) -> _e k n (samples p (pure i))) $ enumerate 
 
 slice :: Pattern Int -> Pattern Int -> ControlPattern -> ControlPattern
 slice pN pI p = P.begin b # P.end e # p
-  where b = (\i n -> (div' i n)) <$> pI <* pN
-        e = (\i n -> (div' i n) + (div' 1 n)) <$> pI <* pN
+  where b = (\i n -> (div' i n)) <$> pI <*| pN
+        e = (\i n -> (div' i n) + (div' 1 n)) <$> pI <*| pN
         div' num den = fromIntegral (num `mod` den) / fromIntegral den
 
 _slice :: Int -> Int -> ControlPattern -> ControlPattern
--- a/src/Sound/Tidal/Core.hs
+++ b/src/Sound/Tidal/Core.hs
@@ -2,7 +2,7 @@
 
 module Sound.Tidal.Core where
 
-import           Prelude hiding ((<*), (*>))
+import           Prelude
 
 import           Data.Fixed (mod')
 import qualified Data.Map.Strict as Map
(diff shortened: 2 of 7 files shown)


tonyday567:
I can see how the use of cycleArcsInArc in pure is appealing but I think it breaks the applicative laws eg

```
-- (identity law) pure id <*> v = v
queryArc (slow "1 2" "a b c" :: Pattern String) (Arc 0 1)
[(0>⅓)|"a",(⅓>½)-⅔|"b",0-(½>⅔)|"a",(⅔>1)-1⅓|"b"]

queryArc (pure id <*> (slow "1 2" "a b c" :: Pattern String)) (Arc 0 1)
[(0>⅓)|"a",(⅓>½)-⅔|"b",0-(½>⅔)|"a",(⅔>1)|"b"]
```

The pure Pattern in this PR, `pure (bipure a v)` is the simpler 'a single event being the value over the query arc'.  Maybe not as useful but it should more easily combine to satisfy the applicative laws.

tonyday567:
This is a better example:

```
 fast 4 saw
(0>1)|0.0
 (fast 4 $ pure id) <*> saw
(0>¼)|0.0
(¼>½)|0.25
(½>¾)|0.5
(¾>1)|0.75
 fast 4 (pure id <*> saw)
(0>¼)|0.0
(¼>½)|0.0
(½>¾)|0.0
(¾>1)|0.0
```

tonyday567:
@yaxu I may be hopelessly confused, but I would guess that `fast 4 saw` should be:

```
(0>¼)|0.125
(¼>½)|0.375
(½>¾)|0.625
(¾>1)|0.875
```

I added a `fast'` function that gave me this.  It pretty much follows `(fast 4 $ pure id) <*> saw` which does a cycleArcsinArc in the pure bit.

I think `fast` is right for Digital patterns but Analog patterns don't work as intended. My understanding of intended might be way off, however.

yaxu:
I don't fully understand what's going on here yet, but some thoughts:

`|*` takes structure from the left (while applying multiplication), so it's surprising that `|*>` takes structure from the right (while not multiplying anything). I'm not sure what to suggest instead, though! `*>|` is not very appealing.

`longArcStop = 1000000` is concerning, that would only be 21 seconds at 48KHz if someone decided to calculate tidal patterns at audio rate, for example. I guess this wouldn't serve your purposes, but could an 'empty arc' more usefully represent an event that doesn't have a start and end, and therefore is continuously 'on'? Or if this is a distinct concept from an empty arc, could we make another constructive for an infinite arc? Would this remove the need for a 'long arc'?

In the past I've wondered whether `pure x` should create an infinitely long event rather than one that fires once per cycle. I guess this is the difference between a pure analogue and a pure digital pattern.

Is it possible that the changes here could make tidal significantly slower? I'd be keen to do some proper benchmarking if so.

tonyday567:
> When you say WIP, does that mean it's just for discussion and not for merging at this point?

Yes, absolutely.  I'm not sure yet whether these changes are a net plus for the library.  I'm pretty comfortable that, if we can, aligning Pattern with the applicative laws will lead to some great code (fast, easy to understand and bug-free), I just don't know whether it's possible yet or if it breaks too many things.

I tend to overshare with a PR and get into the detail.  I got the sense that this is ok here, but if you want an alternative communications channel let me know.  There's a heap of issues already and it might be better to have a more organised way of ticking through them.

yaxu:
@tonyday567 That's all good, just wanted to be sure you weren't waiting on me for things.. I will reply to the other comments eventually though!

yaxu:
Hi @tonyday567 is this train of thought still activer?

tonyday567:
@yaxu, hi! I've lost the thread to some extent, but Tidal is so awesome I'd be happy to pick it up again.  

The main thing I remain confused about is Analog vs Digital.  Is `Pattern Digital q == Pattern Analog (const q)`?

yaxu:
@tonyday567 When querying an analogue pattern, you 'sample' a point along it, taking the halfway time point of the arc:

```
> queryArc sine (Arc 0 1)
[(0>1)|0.5000000000000001]
> queryArc sine (Arc 0 0.5)
[(0>½)|1.0]
```
This is mostly just down to the pattern behaving differently, but the `nature` field is used when combining patterns, so that any structure in the result comes from the discrete pattern(s) input.

yaxu:
Do you have a clear idea about the advantages of the biapplicative work you've contributed? I find the arc/arcf indirection rather difficult to work with so far.

tonyday567:
O dear, it shouldn't get in the way.  ArcF is there just in case you want another concept of time other than rational.  Ratio arithmetic is slooooww.

Do you have an example of where it's confusing things?  Happy to help unwind. Is the main problem the convoluted specification for an Event?

You could do something like this:

```
pattern Event :: Time -> Time -> Time -> Time -> a -> Event a
pattern Event a b c d e = Event' (Arc a b) (Arc c d) e
```

and then an Event looks something like `Event (1%1) (5%4) (1%1) (5%4) 2`.

tonyday567:
@yaxu, to answer your previous question - is this an active train of thought - there is something  fundamental represented in the Pattern and Event types that doesn't exist elsewhere, but development of this is better done outside Tidal.  There's too much going on above this abstraction level.

So I think it's better to close this PR and start from scratch.

PR #459, opened 2019-01-18
Additional Scales and Unit Tests

XiNNiW:
Hi folks,

I took a shot at adding characterization tests for the existing scales used by the scale function. I also added messiaen's modes of limited transposition and provided aliases for the diminished and whole scales to make them easier to find. I think I did everything right and that this is ready to be merged into the main repo but I am a haskell newbie and would appreciate any feedback that folks have.

Thanks,

David


yaxu:
I don't know much about scales but looks great to me!
Sorry about the delay in merging

PR #461, opened 2019-01-23
david minnix -- commiting snowball function for spinning out variatio…

XiNNiW:
Hi all,

I made a function called 'snowball' to provide a quick way to spin out a bunch of melodic variations from a short phrase. It combines a phrase with a transformed version of itself and feeds back a given number of times. This is kind of like putting fx in the feedback path of a delay line. It can do really pretty things to melodies.

For example:

d1 $ (note $ scale "hexPhrygian" $  "0 . 2 3 ~ . 0 1 . -1 -2") #s "gtr"

d1 $ (note $ scale "hexPhrygian" $ snowball (+) (slow 2) 4 "0 . 2 3 ~ . 0 1 . -1 -2") #s "gtr"

Let me know if you have any feedback. I think this is ready to be integrated with the main repo.


XiNNiW:
Hmm.. It seems to have a similar signature but slightly different functionality. stutWith with combines patterns with overlay always and always combines them at a subcycle delay. I am also a bit new to haskell and can't tell for sure if stutWith applies its transform recursively (ie once for the 1st repetition, twice for the second, ect...).

I wrote a test comparing the output and was able to get somewhat close by setting stutWith's delay to a full cycle. However, I couldn't replicate the effect of combining structure from both sides (using an operator like +) again and again like I could with snowball.

Are you thinking there would be a way to unify the functionality, or make snowball part of the stut family of functions?

yaxu:
Sorry for the long delay! Yes I believe `stutWith` does apply the function recursively. I think we need to look at these functions to work out how they relate to each other and then think about the names. I do like `snowball` though!

XiNNiW:
No prob! I took another look at `stutWith` after reading your post. It may be possible to share code with the `stutWith` implementation if the `overlay` function inside it were parameterized.  

As for naming, my initial reaction is that stutter implies layering up/repeating the sound. That part of the effect is up to the caller in snowball. But maybe a name more similar to `stutWith`, a less metaphorical name, might help?

I'll take a shot at making the test pass with an implementation that uses pieces of `stutWith`.

yaxu:
Shall we just merge this as snowball for now? It's a nice name, and having just adding a good number of differently named but kind-of-related functions, I think a completely rational naming scheme for functions is beyond us really.. Maybe we can tackle that one in 2.0!

XiNNiW:
Sounds good! Naming things is one of the hard problems right? Sorry for the delay, btw. Life got busy all of a sudden.  I will take a look at combining this thing with the stut family at some point!

yaxu:
Thanks!

yaxu:
Now released in 1.0.12

XiNNiW:
Cool! I'll write up a doc.

yaxu:
@XiNNiW great! on the wiki?

XiNNiW:
yup! should be up now.

PR #462, opened 2019-01-23
david minnix - implemented soak

XiNNiW:
Hi folks,

I implemented a function called "soak" for repeatedly applying a function to a pattern and combining the results together with cat. This can be a cool way to spin out variations for both melodic and rhythmic patterns. 

For example:
d1 $ (note $ scale "hexPhrygian" $ soak ((0.25 <~).(+ irand(2))) 4 "0 . 2 3 ~ . 0 1 . -1 -2") #s "gtr"

will create a wandering melodic phrase that returns to the original after 4 cycles. 

d1 $ soak (iter 8. slow 2 .ply 3) 4 $ euclid 3 5 $ sound "bd sn"

will build up a kick rush over 4 bars

Let me know what you think. ~~I'm not crazy about the name but I think It could be used for some cool stuff.~~ I changed the name to soak which I think is a better descriptor. 

Thanks,

David

### DIFF
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -1727,12 +1727,12 @@ smooth p = Pattern Analog $ \st@(State a cm) -> tween st a $ query monoP (State 
 swap :: Eq a => [(a, b)] -> Pattern a -> Pattern b
 swap things p = filterJust $ (\x -> lookup x things) <$> p
 
-{- @mutateBy@ | 
+{- @coat@ | 
     applies a function to a pattern and cats the resulting pattern,
     then continues applying the function until the depth is reached
     this can be used to create a pattern that wanders away from 
     the original pattern by continually adding random numbers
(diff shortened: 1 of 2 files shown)


yaxu:
Nice! What do you think about swapping the parameters? Depth comes first for `stut`/`stutWith`, and I find it clearer to put the 'least complicated' parameter first - in your examples the depth parameter gets kind of lost after the function.

XiNNiW:
Sounds great! I'll switch them around and commit again.

XiNNiW:
Hey alex, I think this one might be ready to go. The parameters have been swapped to put depth first.

Now it looks like this in use:

`d1 $ soak 4 (iter 8. slow 2 .ply 3) $ euclid 3 5 $ sound "bd sn"`

yaxu:
Thanks, will merge! We might tweak the name later to create a family of functions but I like the name soak..

PR #465, opened 2019-01-25
david minnix - saw that yaxu had already fixed the bug but added a un…

XiNNiW:
…it test to prevent regressions

Hi,

I saw the bug description and started this branch to take a shot at fixing. Then I saw that @yaxu had already fixed it! I had already written a unit test to check for the bug. It fails appropriately when the fix to sometimesBy is removed. I though I might as well open a PR as the test might be useful for preventing regressions. 

Cheers,

David


yaxu:
A nicely written test, thanks a lot!

PR #466, opened 2019-01-26
patch parse error on emacs init

annybs:
multi let emits "parse error on input '='" for every definition and emacs repl fails to initialise, letting each value individually is more robust

I'm aware c/o @yaxu that this is fixed in BootTidal.hs which presumably resolves the bug for other environments, but emacs still uses the tidal.el which I guess was outdated.

Perhaps a 'better' fix would be to read the BootTidal.hs file directly from disk? I don't know Lisp though, so no idea how to do that.


yaxu:
Thanks!

PR #467, opened 2019-01-27
a fix to unwrap' so fit' will work again

bgold-cosmos:


yaxu:
What do you think about the above, @bgold-cosmos ? I can have a look at these things if you like

bgold-cosmos:
OK, looking at this more I _think_ the `unwrap'` that `fit'` has been using is actually the same as `unwrapSqueeze`.  The implementations are rather different, though, so I'm going to try thinking of some weird test cases to see if there's any difference.  If not, then I can just get rid of `unwrap'` and have `fit'` refer to `unwrapSqueeze`.

yaxu:
Aha, well spotted!

Ok so thinking about the name.. `unwrap` is actually `join` (except we can't directly define it as that due to the way monads are structured in haskell or something). It uses structure from both inner and outer patterns.

`innerJoin` only uses structure from inner patterns, and `outerJoin` from outer patterns.

`unwrapSqueeze` is like `unwrap`/`join` in that it uses structure from both patterns, but with a different strategy.. 

Well, I can't think of a better name for this than `joinSqueeze` at the moment.

Btw `unwrapSqueeze` (or `unwrap'`) isn't used very much at the moment, I think there is likely some untapped functionality here in patterning patterns in interesting ways..

bgold-cosmos:
OK, I've updated this and tested it out.  `unwrapSqueeze` and `unwrap'` were indeed identical in intent. So I did a few things:

  * fixed a big in unwrapSqueeze
  * renamed unwrapSqueeze to squeezeJoin
  * removed the redundant unwrap'
  * updated functions using unwrap' to refer to squeezeJoin

yaxu:
Looks great, thanks very much!

PR #470, opened 2019-02-01
david minnix -- implemented function for listing out all of the chord…

XiNNiW:
…s tidal can make in a pattern

Super simple function. I found the one for listing scales super useful so I made one for listing the chords.


XiNNiW:
Good point! I did it out habit. (test driving is pretty drilled into me) I'll delete it.

XiNNiW:
Good point! I did it out habit. (test driving is pretty drilled into me) I'll delete it.

XiNNiW:
Gone!

yaxu:
Thanks a lot! I guess the test wasn't doing any harm, and I'm not fully up to date with testing philosophies, but hey!

XiNNiW:
No prob! To me, it depends on what your building. I think your right. For someone just trying to add a chord to tidal, the burden of changing it in two places outweighs the benefit of increased documentation (the function is very simple) and early warning system for issues. (like you said, not much logic in there!)

PR #471, opened 2019-02-01
david minnix - fixing Issue 468

XiNNiW:
Hi, 

I took a shot at fixing Issue 468. I made it so 0 and 1 are special cases of wedge. 0 results in the original pattern, 1 in the replacement pattern. Hope that's right.

Thanks!

David


yaxu:
Thanks a lot!

PR #473, opened 2019-02-08
implements onceFor and asapFor

bgold-cosmos:
also I believe this fixes a bug where `once` was actually behaving like `asap` and not aligning to the currently playing cycle boundary


yaxu:
That was the intended behaviour for `once`. `asap` and `once` are the same except `asap` ignores 'artificial latency' so if you send a cycle full of events, the timing at the start might not align. 
In practice, I don't know if it's possible to tell the difference.. So your idea about what `once` should do might well be better!

OscarSouth:
Commenting on the difference between `asap` and `once` -- intuitively and in terms of the expectation of these functions in performance, I would say that the functionality suggested by @bgold-cosmos sounds more useful to me. I personally expected `asap` to start sending the pattern as soon as possible (and I think a minutia latency isn't an issue -- I've not been able to tell much difference between `once` and `asap`) and `once` to start on the next sync quantum. I feel that this will match with most peoples user expectation.

To keep all the current functionality there, perhaps an `asap'` which functions is the `asap` of today, then `asap` could function as `once` does today and `once` could align to the next sync quantum. This is the kind of behaviour I would have intuitively expected.

yaxu:
Have thought a bit about this:

Jitter is probably a greater enemy than latency, so lets dump the original `asap` functionality.

I know it was my idea, but `onceFor` is a bit of an oxymoron, `onceFor 2` kind of means `twice`. We could do with better names here..

`once` (and the now doomed `asap`) were really intended for single-fire events like `once $ s "airhorn"`. After some thought I decided to send a cycle's worth, but it's not really ideal.. It sends all the events for that cycle at once, with timestamps in the future.. If there's a lot, that might cause problems.

The proposed new `once` and `onceFor` seem closely related to the `mortal` transition. I think a transition is better here for the above reason.. A transition will be calculated at the right framerate rather than as a whole cycle or more.

So my proposal is..

- make `asap` an alias of the existing `once`. 
- define a new transition that's like mortal but simpler - just one parameter for how long the pattern should live for. Perhaps it could simply be called `for`.

I'm not 100% on the second proposal.. It is nice to be able to trigger things without having to give a pattern id.. But then again if you said `for 100 $ sound "bev"` when you meant to type `for 10 $ sound "bev"`, you'd be thankful of a way to stop that from playing...

echophon:
I think the first proposal is fine.

On the second, I wonder if there would be confusion naming it `for`?  `while` or `until` may be alternatives.  What if it was still a named transition, so that `silence` could still be used in the above `"bev"` situation?  I could see this being very useful to add small temporary patterns.  Maybe even an additional parameter as an offset starting cycle.  It would end up being very similar to `seqP` in this case but might be good to have an alternative.

```
do
  d1 $ s "bd*4"
  while "rampup" 4 0 $ s "hh*16" # gain (slow 4 $ saw)  
  while "rampdown" 4 4 $ s "hh*16" # gain (slow 4 $ rev saw)
```

bgold-cosmos:
I'll take a look at a transition version, since I think the scheduling issue is important.  Is it possible to have a transition that keeps playing the previous pattern _and_ doesn't add the new one to the stack?  That might be an ideal behavior for something like `once`.  I'm pretty sure the former is easy, not so sure about the latter.  If the new "temporary" pattern gets put on the stack then a later `xfade` might behave oddly...

bgold-cosmos:
I've made a new PR with a transition version #474 which I think should overall be more flexible, so unless someone objects I can close this one.

PR #474, opened 2019-02-12
`onceFor` and support for "momentary" transitions

bgold-cosmos:
`onceFor id t pat` now works to overlay a pattern for a limited time on the given PatID.

To make this work, there's a new function `momentaryTransition` which works much like `transition` but doesn't update the history.  This keeps future transitions from suddenly reviving the momentary pattern.

In the process I also discovered and fixed two minor bugs in Stream.hs
  * the first time streamReplace is used it wasn't properly adding the pattern to the history
  * hush was silencing all playing patterns but not changing the history, which makes future transitions behave unexpectedly


bgold-cosmos:
I also want to add that while I believe this PR works correctly, there appears to be a SuperDirt issue (https://github.com/musikinformatik/SuperDirt/issues/109) that can make it seem like it's not quite right, but the problem very much seems to be on the SuperDirt end, not Tidal.

yaxu:
Thanks for this, @bgold-cosmos !

Continuing the discussion from the previous PR, is there a difference in behaviour between `onceFor 0 2` and mortal 0 2 2` ?

I think we also need a better name than `onceFor`. I suggested `for`, for example.. But while `for 10 pattern` is quite clear, the pattern id gets in the way a bit in `for 2 10 pattern`. Howabout we have two versions, a `for` that doesn't use an id (i.e. so successive calls to `for` don't overwrite each other), and something like a `forId` that takes a pattern id as its first parameter?

yaxu:
Thanks for this, @bgold-cosmos !

Continuing the discussion from the previous PR, is there a difference in behaviour between `onceFor 0 2` and mortal 0 2 2` ?

I think we also need a better name than `onceFor`. I suggested `for`, for example.. But while `for 10 pattern` is quite clear, the pattern id gets in the way a bit in `for 2 10 pattern`. Howabout we have two versions, a `for` that doesn't use an id (i.e. so successive calls to `for` don't overwrite each other), and something like a `forId` that takes a pattern id as its first parameter?

bgold-cosmos:
They are somewhat similar, but:
  * `mortal` immediately switches to the new pattern, then fades it out after a time
  * `onceFor` overlays the old and the new, then abruptly terminates the new (and the old keeps playing)

Internally, `mortal` uses `transition` while `onceFor` uses `momentaryTransition` to avoid adding the new pattern to the history.  So the upshot of all this is that if you do:

```
d1 $ s "bd sn:2"

-- wait a bit

onceFor 1 1 $ s "rave"

-- wait a bit

xfade 1 $ s "cp*4"
```
you'll hear the kick-snare pattern, the rave sample will play once on the downbeat after the `onceFor` is evaluated, but the kick-snare will continue playing after that.  And then the final xfade will smoothly fade from the kick-snare to the claps, as if the rave sample never happened.

yaxu:
Aha, I see. What's the benefit of two patterns sharing a pattern id in this way? That seems a confusing, I'd expect this to replace an existing pattern with the same id.

How about
* `for n $ s "bd sn"` for running an 'anonymous' pattern for `n` cycles, where there's no way to stop/replace it (apart from 'hush', maybe)
* `forId i n` for running a pattern with an id, so that it can be replaced while it's ongoing

I think the pattern probably should still be added to the history in the latter case, even though it will go silent after the given number of cycles.. You'll just be transitioning from silence, then.

telephon:
In supercollider, we postfix such functions with `def`, and usually it takes a name not only a number. In tidal, maybe there could be a function `def :: id -> (a -> b) -> (a -> b)` that takes a transition function and "identifies" it. Then you could write `def i for n`, couldn't you?

bgold-cosmos:
> I think the pattern probably should still be added to the history in the latter case, even though it will go silent after the given number of cycles.. You'll just be transitioning from silence, then.

That's not how transitions generally work, though.  For example:

```
tidal> d1 $ s "bd sn:2" # cps 1
-- kick snare
tidal> mortal 1 2 2 $ s "cp*4"
-- sharp transition to claps, which then end
tidal> xfade 1 $ s "hh*8"
-- claps suddenly come back, then fade to hihats
```
The last part is because the only thing that goes on the history stack is the pattern `s "cp*4"`, not any info about it ending - that was handled by the transition itself.  So the claps get "revivified" by any subsequent transition.

Maybe that's not the intended behavior and transitions should be reworked a bit, but the way I implemented `onceFor` was driven by the necessity of working around that behavior.

yaxu:
Hmm yes, that seems.. not too good. I implemented transitions without really using them properly, so there probably are UI issues.

I can remember why I implemented it that way though, adding transitions on top of transitions all the time also seems like a bad idea, in terms of efficiency.. Hmm!

bgold-cosmos:
The particular way I dealt with it was to keep the old generic `transition` function but make a new one `momentaryTransition` that is almost the same but doesn't add to the history.  These could probably be merged into a single function with a "momentary" flag as another argument.  Both types of transition seem useful.

bgold-cosmos:
I've made a few changes in response to suggestions here (and in the related #473 discussion).

  * asap = once
  * onceFor renamed to forId
  * streamOnce simplified to remove now-pointless asap code
  * some other code cleanup

I did not implement `for` without a pattern ID because then there's no way to hush it, as @yaxu pointed out in #473.  The same is true for `once` but since that's always a single cycle that seems like less of a problem.

Right now `forId` is set up so that the given pattern will start the next time the cycle count is at the beginning of the given duration.  For example, if you use
`forId 1 4 $ s "<bd sn hh cp>"`
but the current cycle count is 1, Tidal will wait until cycle 4 so that the pattern starts playing on the "bd".  This can be useful.

OTOH, it means if you use `forId 1 8 $ s "hh*8"` you may have to wait a while before you actually hear anything, and in this case the user probably doesn't care whether the "hh*8" starts on a cycle that's a multiple of 8.

I'm not really sure if one is better than the other, or maybe both should be provided?  I think I'd lean toward the latter, but I'm not sure about names.  Maybe `syncForId` for the first version?

PR #478, opened 2019-02-21
add support for new spectral fx

madskjeldgaard:
Add new parameters to be used with the SpectralTricks fx that are now added to SuperDirt by default, see [superdirt pull request](https://github.com/musikinformatik/SuperDirt/pull/113)

### DIFF
--- a/src/Sound/Tidal/Params.hs
+++ b/src/Sound/Tidal/Params.hs
@@ -385,6 +385,44 @@ ringdf = pF "ringdf"
 distort :: Pattern Double -> ControlPattern
 distort = pF "distort"
 
+-- Spectral conform
+real :: Pattern Double -> ControlPattern
+real = pF "real"
+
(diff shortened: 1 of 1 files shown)


src/Sound/Tidal/Params.hs
```diff
@@ -385,6 +385,44 @@ ringdf = pF "ringdf"
 distort :: Pattern Double -> ControlPattern
 distort = pF "distort"
 
+-- Spectral conform
```
telephon: it seems that not all parameters are here. At least `xsdelay` and `tsdelay` are missing?

madskjeldgaard: You're totally right. I must have fucked up and accidentally deleted xsdelay, tsdelay and freeze. Will fix

madskjeldgaard: Fixed

yaxu:
Thanks!

PR #482, opened 2019-02-24
Change tidal.el to use BootTidal script

lnfiniteMonkeys:


yaxu:
Does this require the BootTidal.hs script to be in the same directory as the one you started emacs in?

lnfiniteMonkeys:
Not really, since the string is being sent to the interpreter, so as long as the interpreter can find it it's fine.

lnfiniteMonkeys:
That being said, I just realised that the interpreter can only know of it if it's pointed to the full path, launched in the same directory, or `:cd`ed there.

Both of these should work the same:
```elisp
(tidal-send-string (concat ":cd " tidal-path))
(tidal-send-string ":script BootTidal.hs")
```

```elisp
(tidal-send-string (concat ":script " (file-name-as-directory tidal-src) "BootTidal.hs")
```
where `(defvar tidal-path "~/tidal")` or whatever the default is, which can be changed by the user.

The former has the benefit of consistency, since every GHCi process will have the same working directory regardless of where it was started. 

The current directory in the latter remains the one that the interpreter was launched in (the directory of the `.tidal` buffer that was used to start), which may or may not be desirable.

yaxu:
I see, I think I prefer it as it is, so no configuration is necessary by default.

Ideally though a `BootTidal.hs` in a standard location (e.g. `~/.local/share/tidal/BootTidal.hs`, or the windows equivalent) would be loaded if present, or a script in a different location could be loaded via a configuration setting.

lnfiniteMonkeys:
No configuration would be necessary if tidal is located in the home directory, and if not adding this line to the `.emacs` or `init.el` should point it accordingly:
```elisp
(setq tidal-path "path/to/tidal")
```

I have also found it useful in general to have a reference to the source directory, so in addition to importing the installed module you can also compile and load straight from the source. This also means you can be experimenting with multiple versions/forks in different directories without getting the packages mixed up, or have to mess with the installed package when you want to try something.

Do you expect users to have cloned the source repo, or is it more of a "just run `cabal install tidal` and fire away" thing?

Perhaps an installation script could be used instead of installing directly from the terminal? It could copy `BootTidal.hs` to a default location where it can be found by all editors/interpreters, so then it wouldn't even matter if or where the source is.

Then the `tidal.el` file could have `(defvar tidal-boot-script-path "path/to/boot/script")`, and that could be overwritten by the user to point to a custom boot script as above.

yaxu:
How about by default, interrogating cabal/stack for the data-dir for the tidal package and looking in there?

lennart:
@lnfiniteMonkeys since I wanted to have this as well, I went along and tried to implement what @yaxu suggested, see #495 

regarding your question: 
I guess we expect a `cabal install tidal` (I added BootTidal.hs into `data-files` so it will be distributed to a common location)

yaxu:
Excellent, thanks both!

PR #485, opened 2019-03-02
Topic/internal signal routing

telephon:
These are three parameters which I need for a new signal routing schema in superdirt. Given the right synthdefs, you should be able to do the following:

```haskell
d1 $ # sound "cp*4" # to "0"
d2 $ # sound "mySynth" # from "0" # toArg "soundIn"
```

Then the `soundIn` parameter of `mySynth` should receive the signal from `d1`.

I'm not sure if this PR is correct, because I haven't been able to set up a combination of cabal and github yet, so I couldn't test.


bgold-cosmos:
This looks very interesting!  I'm anxious to test as soon as I'm able.

PR #495, opened 2019-03-19
load BootTidal.hs in emacs without manual config

lennart:
infers path to BootTidal.hs and loads it in `tidal.el`

this way we only have to maintain one bootscript.

this works as long as `ghc-pkg` can be found in path and tidal is installed via `cabal install` in the global package database.

however, you can point to you custom `BootTidal.hs` via
```
(setq tidal-boot-script-path "~/MyTidalBoot.hs")
```

In order for this to work out-of-the-box we have to add the `BootTidal.hs` as a data-file in `.cabal`.

We try to infer the path of the installed tidal by asking for the data-dir of the latest installed version of tidal.

this duplicates #482 but should merge cleanly into current master


yaxu:
Thanks!

PR #505, opened 2019-04-04
Add explicit Ord and Eq instances for the Value type

bgold-cosmos:
With this change it's now possible to sensibly compare Values of different types.  VS are always less than the numeric types, and
```
tidal> VF 1 < VI 2
True

tidal> VI 1 < VF 2
True
```

This also means expressions such as `(>=) <$> n "1 3" <*> n 2` or even `(>=) <$> n "1 3" <*> 2` gives `(0>½)|False   (½>1)|True` as one would expect.

When using `fmap` in such a way it's still possible to uncover some strange situations when trying to compare maps with different sets of keys.  I don't know if it's possible to resolve all of those -- for now I leave it for another PR :)


yaxu:
Thanks!

PR #519, opened 2019-05-23
Initial commit of Markov chain tools

bgold-cosmos:
`runMarkov n tmat xi` generates a Markov chain (as a list) of length `n`
using the transition matrix `tmat` starting from initial state `xi`.
Each entry in the chain is the index of state (starting from zero). 
Each row of the matrix will be automatically normalized. For example:
```
runMarkov 8 [[2,3], [1,3]] 0
```
will produce a two-state chain 8 steps long, from initial state 0, where the
transition probability from state 0->0 is 2/5, 0->1 is 3/5, 1->0 is 1/4, and 
1->1 is 3/4.

`markovPat n xi tp` generates a one-cycle pattern of `n` steps in a Markov
chain starting from state `xi` with transition matrix `tp`. Each row of the
transition matrix is automatically normalized.  For example:
```
tidal> markovPat 8 1 [[3,5,2], [4,4,2], [0,1,0]]

(0>⅛)|1
(⅛>¼)|2
(¼>⅜)|1
(⅜>½)|1
(½>⅝)|2
(⅝>¾)|1
(¾>⅞)|1
(⅞>1)|0
```

### DIFF
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -1051,7 +1051,7 @@ rationals are required) -}
 lindenmayerI :: Num b => Int -> String -> String -> [b]
 lindenmayerI n r s = fmap (fromIntegral . digitToInt) $ lindenmayer n r s
 
-{- | @runMarkov@ n tmat xi@ generates a Markov chain (as a list) of length @n@
+{- | @runMarkov n tmat xi@ generates a Markov chain (as a list) of length @n@
 using the transition matrix @tmat@ starting from initial state @xi@.
 Each entry in the chain is the index of state (starting from zero). 
 Each row of the matrix will be automatically normalized. For example:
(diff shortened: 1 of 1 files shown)


yaxu:
Looks nice, what do you think about each cycle having a different random seed? It would be a lot less efficient - each query would need to calculate a whole cycle's worth of events (with splitQueries applied to deal with queries over more than one cycle), but maybe a bit more tidal-ish?

bgold-cosmos:
Yes, interesting idea.  If ghc is somewhat smart about caching it might not even be a big CPU hit.  I'll see if I can get a test version going.

bgold-cosmos:
Here's a new version, I like it much better.  The old behavior can always be recreated with `loopFirst`

yaxu:
Nice! It might be worth adding this to the new profiling stuff.

PR #521, opened 2019-06-04
modified to the implementation of chooseBy to use !!!

XiNNiW:
When playing around with `chooseBy` I found it hard to keep the index in bound so I took a shot at making so the input would be autoscaled using `!!!`.


yaxu:
A single character but very useful change, thanks!

PR #523, opened 2019-06-18
Variable probability with '?'

tuchella:
Allows specifying the degradeBy value like this `d1 $ s "bd?0.3 sd?0.8"`. Without number also still works `d1 $ s "bd?"`

And added some additional pattern parser tests


yaxu:
Looks good but wanted to check that it works with patterns of floats as well? E.g. I think `"0.4 0.5? 0.6"` shouldn't be interpreted as `"0.4 0.5?0.6"`

yaxu:
Looks perfect, thanks!

PR #524, opened 2019-06-23
define arpeggiate with arpWith

nini-faroux:
just avoids some duplication, had to change the arpWtih function slightly by adding the 'sortOn whole es' bit to munge, so feel free to ignore this if that's an issue.


yaxu:
Good call, thanks a lot!

PR #528, opened 2019-06-28
Add variable interpolation in patterns

bgold-cosmos:
Adds a new function `substP` which enables variable interpolation in patterns, used like this:

```
d1 $ s (substP ["bd", "sn cp"] "$0 $1")
```
is equivalent to
```
d1 $ s "[bd] [sn cp]"
```
the advantage of this over similar functions such as`fit` is that it works by string substitution, so the interpolated variables can be any valid pattern.  For example:
```
d1 $ s (substP ["bd(3,8)", "<sn:2 cp>"] "<$0 $1>")
```

The substitution is cyclic so
```
d1 $ s (substP ["bd(3,8)", "<sn:2 cp>"] "<$0 $1 $2 $3>")
```
just sounds the same as the previous example.

At the moment only single digits are allowed.  A `$` followed by something other than a digit will be eliminated.

Comments and suggestions welcome!

### DIFF
--- a/src/Sound/Tidal/ParseBP.hs
+++ b/src/Sound/Tidal/ParseBP.hs
@@ -474,6 +474,24 @@ pRatio = do s <- sign
 pRational :: Parser (TPat Rational)
 pRational = TPat_Atom <$> pRatio
 
+substP :: (Enumerable a, Parseable a) => [String] -> [Char] -> Pattern a
+substP subs p = toE $ runParser sub subs "" p where
+  toE (Left e)   = E.throw $ TidalParseError e p
+  toE (Right tp) = parseBP_E tp
+  sub :: GenParser Char [String] String
+  sub = concat <$> many (
+          (many1 $ noneOf "$")
+          <|>
+          try ( do char '$'
+                   d <- read <$> sequence [digit]
+                   s <- getState
+                   return $ if (length s > 0)
+                            then "[" ++ s!!(d `mod` length s) ++ "]"
+                            else ""
+              )
+          <|> (char '$' >>  return "")
+          )
+
 {-
 pDensity :: Parser (Rational)
 pDensity = angles (pRatio <?> "ratio")


bgold-cosmos:
The major downside to this is that because it works as string interpolation rather than starting with patterns, you can't do things like
```
d1 $ s (substP [fast 2 $ "bd(3,8)", "<sn:2 cp>"] "<$0 $1 $2 $3>")
```

I've been trying to get a "true" interpolation going, but AFAICT the main way to do this is to make the entire Tidal parser stateful, which gets difficult because of the polymorphism.

yaxu:
Have you looked at `inhabit`? It's undocumented but I think it's similar

bgold-cosmos:
Ah OK, thanks for reminding me of that.  Playing with it a bit I think the missing feature that I really want is the ability to mix "lookups" with an ordinary pattern.  For example, I'd love for this to work:
```
d1 $ s $ inhabit [("a0", "bd sn:2"), ("b1", "cp*4")] "a0 hh b1 hh"
```
where `hh` would just refer to the normal hihat sample folder.  But it may be possible to modify `inhabit` to do something like that - I'll try playing around a bit.

bgold-cosmos:
Easier than I expected, this basically does what I wanted:
```
inhabit' :: (Enumerable a, Parseable a) => [(String, Pattern a)] -> Pattern String -> Pattern a
inhabit' ps p = squeezeJoin $ (\s -> fromMaybe (parseBP_E s) $ lookup s ps) <$> p
```
In practice this can get a little strange if the lookup keys resemble normal pattern symbols, but that might be more easily avoided with some helper functions:
```
a $$ b = inhabit' (zip (map show [0..9]) a) b
a $# b = inhabit' (zip (map (\x -> [x]) "hijklmnopq") a) b

d1
 $ s (["bd sn:2", "cp(3,8)", "hc(3,8)"] $$ "0 <2 gabba> <1 0> <2 1 0>")

d2
 $ n ([run 8, "0 1 0 4"] $# "h c*2 i c*2")
 # s "supertron" # legato 0.3
```

yaxu:
Template haskell could be a good way forward for this use case

PR #531, opened 2019-07-02
MiniTidal haskell-src-exts refactor

dktr0:
MiniTidal has been refactored to use haskell-src-exts. The impacts of this refactor are twofold: (1) many idioms involving operators, left sections, right sections are now supported (or supported "correctly", eg. with correct fixity), and (2) it will be easier to maintain and extend moving forward. It adds a haskell-src-exts dependency for which bounds based on the latest stackage LTS for ghc versions from 7.10.3 to 8.6.3 have been set. With this pull request, MiniTidal supports everything in the main tutorial page at tidalcycles.org with the exception of transitions (currently out of scope for something which parses to ControlPattern).


dktr0:
Is there something broken in the Travis testing system? I notice that it is failing because I have set bounds on haskell-src-exts that are appropriate to the LTS for each ghc version - it wants to pick the ultra newest version (eg. even for GHC 7.10.3!). Any thoughts?

yaxu:
It's just an extra check I added to keep us aware of new versions so we can increase the bounds appropriately (after testing / fixing as required).

dktr0:
Oh I see. So better to raise the bound to allow the new haskell-src-ext then?

yaxu:
Yep!

yaxu:
As much as I like typing `cabal install happy alex`, I think these dependencies of haskell-src-exts are too much, requiring extra install steps which won't be of any benefit to people who aren't using minitidal. Howabout we put minitidal in a separate hackage package, but still keep everything in the same repo, so they can continue to be tested together by Jenkins. Any drawbacks @dktr0?

dktr0:
I think that sounds fine. My only pause is about the name - perhaps it shouldn't be called MiniTidal anymore since it is not really very Mini anymore and is getting less Mini every week (I think next week I will do the refactor so that "everything" ie. things like t1 (xfadeIn 4) $ ... etc is parsed by [MiniTidal] (and then, elsewhere, that all that works in Estuary). So I feel like it needs a more descriptive name. tidal-parser (although it invites a bit of confusion with the parsers "within" the main Tidal package) comes to mind?

dktr0:
(Basically, the moment of making a new hackage package seems like the right moment to change the name.)

yaxu:
maybe `tidal-parse`?

dktr0:
Works for me!

dktr0:
We can rename the modules Sound.Tidal.Parse.*

PR #540, opened 2019-07-15
Add support for inversions to chord parser:

bgold-cosmos:
`"c'major'i"` will give [4,7,12]
`"c'major'ii"` will give [7,12,16]
`"c'major'4ii"` will give [7,12,16,19]


yaxu:
Nice! and now I know what an inversion is

PR #556, opened 2019-10-16
Build emacs tidal boot path in more cross platform way

ethancrawford:
👋 Nice to meet you! My first Tidal PR. 
I've read the contributing and code of conduct docs, and I think I've done everything correctly, but let me know if there's anything amiss.
This PR:
Build tidal boot path in more cross platform way
Previously Windows platforms were not handled. This change allows the BootTidal
script to be found there also.

I'm fairly new when it comes to lisp/elisp, (and windows cmd/batch file programming for that matter) so the code in this PR may not be the same as some written by someone more experienced with it; I'm happy to take on board any comments anyone may have 🙂 

### Testing:
I have the Tidal code checked out locally on both Windows and macOS computers. I started Tidal from emacs in both of them using the code in this change successfully.

### DIFF
--- a/tidal.el
+++ b/tidal.el
@@ -58,12 +58,22 @@
   "*Arguments to the haskell interpreter (default=none).")
 
 (defvar tidal-boot-script-path
-  (concat (substring
-           (shell-command-to-string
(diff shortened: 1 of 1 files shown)


yaxu:
Looks great, thanks very much!

yaxu:
It seems to work fine under Linux too, by the way

PR #576, opened 2019-12-01
Parse updates

yaxu:
Adding some functions as I find them missing..


yaxu:
(sorry about the irrelevant tidying of Tempo.hs there..

PR #581, opened 2019-12-03
Store code position alongside event values

yaxu:


dktr0:
I see you are backing out of the source position in tidal-parse - note that I have a pull request into the emit branch that makes the source position work with Haskellish.

yaxu:
@dktr0 yes I was just getting it out of the way so we had a branch that compiled, as I didn't think it was useful for you and was thinking of merging it into master in the interim. I didn't spot the PR, will have a look ta!

yaxu:
I had to strip the source code context from the events in the tidal-parse tests, but I think it's ready to merge into master now.

dktr0:
We're just pushing and pulling things in all kinds of directions here!

PR #582, opened 2019-12-03
Emit and tidal-parse

dktr0:
tidal-parses uses Haskellish (rework of former ExpParser) and parseBP in tidal-parse gets source location info that way. Still to do (soon) is to release Haskellish as a proper hackage module, but in the meantime it works if you clone the repository at https://github.com/dktr0/Haskellish.git and do cabal install inside that folder.


dktr0:
haskellish is on hackage now http://hackage.haskell.org/package/haskellish-0.1.0

PR #585, opened 2019-12-05
Infer BootTidal.el path on OSX with new Cabal 3

### LINKED ISSUE #583: tidal.el fails to infer `tidal-boot-script-path` on new GHC and Cabal
**Describe the bug**
The Emacs mode tidal.el fails to infer where BootTidal.el is, and so `C-c C-s` leads to error in the Haskell comint window rather than a successful connection with SuperCollider.

**To Reproduce**
1. Get SuperCollider running.
1. Press `C-x C-f` `where-were-you-in-92.tidal`.
1. Tidal mode loads and all seems to be well.
1. Press `C-c C-s` (`tidal-start-haskell`) in the hopes of connecting to SuperCollider.
1. *Variable not in scope* errors for *ghc*, *pkg* and others.

**Expected behavior**

Basically 

**Screenshots**

The following Comint window between Emacs and Haskell, with errors.
```
GHCi, version 8.8.1: https://www.haskell.org/ghc/  :? for help
Loaded package environment from /Users/maco/.ghc/x86_64-darwin-8.8.1/environments/default
Prelude> syntax:  :script <filename>
Prelude> 
<interactive>:2:1: error: Variable not in scope: ghc

<interactive>:2:5: error: Variable not in scope: pkg

<interactive>:2:10: error: Variable not in scope: command :: [a]

<interactive>:2:18: error:
    • Variable not in scope: line :: t0 -> ([Char] -> a0) -> [a]
    • Perhaps you meant ‘lines’ (imported from Prelude)

<interactive>:2:23: error: Variable not in scope: syntax
Prelude> 
<interactive>:3:27: error:
    • Syntax error on 'ghc
      Perhaps you intended to use TemplateHaskell or TemplateHaskellQuotes
    • In the Template Haskell quotation 'ghc
Prelude> 
```

**Environment (please complete the following information):**
 - OS: OSX
 - GHC version 8.8.1
 - Tidal version 1.4.5
 - SuperCollider version 3.10.3

**Additional context**
I looked into `tidal.el` and what is going on there. I searched for those two variables which are out of scope, namely *ghc* and *pkg*. This led me to investigate the value of `tidal-boot-script-path` which surprisingly has the following, 3-line value:

    "ghc-pkg: cannot find package tidal
    ghc-pkg: command-line syntax error
    For usage information see 'ghc-pkg --help'./BootTidal.el

Ok so the lines 60 to 76 try to infer where `BootTidal.el` recides.

https://github.com/tidalcycles/Tidal/blob/bdbe6ee9e43ed7f66d79432522e85a99c1dc4d4e/tidal.el#L60

```elisp
(defvar tidal-boot-script-path
  (let ((filepath
         (cond
          ((string-equal system-type "windows-nt")
           '(("path" . "echo off && for /f %a in ('ghc-pkg latest tidal') do (for /f \"tokens=2\" %i in ('ghc-pkg describe %a ^| findstr data-dir') do (echo %i))")
             ("separator" . "\\")
             ))
          ((or (string-equal system-type "darwin") (string-equal system-type "gnu/linux"))
           '(("path" . "ghc-pkg describe $(ghc-pkg latest tidal) | grep data-dir | cut -f2 -d' '")
             ("separator" . "/")
             ))
          )
         ))
    (concat (substring (shell-command-to-string (cdr (assoc "path" filepath))) 0 -1) (cdr (assoc "separator" filepath)) "BootTidal.hs")
  )
  "*Full path to BootTidal.hs (inferred by introspecting ghc-pkg package db)."
)
```

This inference fails, because apparently new GHC and Cabal do not find installed libraries, as things have changed over there (see https://gitlab.haskell.org/ghc/ghc/issues/17341). Using the `-f` switch to point Cabal to *package.db* is necessary, and I can verify this because `ghc-pkg list tidal` says Tidal is not installed, but `ghc-pkg list -f ~/.cabal/store/ghc-8.8.1/package.db tidal` tells me, correctly, that Tidal 1.4.5 is indeed installed.

The documentation for ghc-pkg's `-f` is

>   -f FILE/DIR    --package-db=FILE/DIR    use the specified package database

Ok now with the new wisdom, in Emacs if I `M-: (setq tidal-boot-script-path "/Users/maco/.cabal/store/ghc-8.8.1/tdl-1.4.5-64a02384/share/BootTidal.hs")` and then `C-c C-s` in my tidal-mode buffer, it is ravetime.

Sorry I do not even understand basically at all how GHC, Cabal and Stack keep track of my packages, what gets installed where and it is all a mess to be honest. However, the above documents both the issue I encounter and provides a solution which works *for me*, namely manually setting *tidal-boot-script-path* to point to my BootTidal.hs.

I see the following options.

1. Add `-f ~/.cabal/store/` *GHC version here* `/package.db` to line https://github.com/tidalcycles/Tidal/blob/bdbe6ee9e43ed7f66d79432522e85a99c1dc4d4e/tidal.el#L68
1. Have `ghc-pkg` without `-f` and an argument.
1. Redesign the `tidal-boot-script-path` inference on lines 60-76 to not rely on ghc-pkg.

This Cabal stuff is being discussed, in general terms, also in #572.

xmacex:
Partially fixes #583, on OSX machines which have Cabal 3.*, GHC 8.8.1 and Haskell packages under user directory. A sketch for someone to expand on, maybe particularly dealing with situations where Tidal is not installed under home directory. Also this fix has no backward compatibility which sucks, sorry but I hope this is useful or motivational for someone :)


yaxu:
I think this should work for old and new cabal:
```
ghc-pkg field tidal data-dir
```

xmacex:
Nice catch that `field` command of ghc-pkg.

I presume the Windows side is currently broken in this PR, until https://github.com/tidalcycles/Tidal/blob/facf1df7ef8e562bbbb28f886a6aaff69beef2e5/tidal.el#L64 returns `data-dir: *the actual data dir here`.

Writing this, I realize that my modifications are making the semantics of the variable `path` drift; in my edits, it is not actually a path, but a `key: value` pair, which get's parsed to a path a few blocks out, a few lines later. Should it be renamed from `pkg-data-dir-variable`, or alternatively push the `shell-command-to-string` into the `(let (filepath))`?

Lol micromanagement/thinking-aloud (sorry). This is my first ever Emacs Lisp contribution to anything ever.

yaxu:
Thanks, nice elisp! (I have no idea about elisp ;) )

I think e.g. the atom plugin just runs  `ghc-pkg field tidal data-dir`, without the `-f` path. Then it works for both windows and mac and we can get rid of that o/s test. Do you think the `-f` is needed?

yaxu:
Ok tried it under linux and -f is needed, at least if cabal is installed with the new default install method.

It also needs `(require 'subr-x)` at the top (I guess you have something else loading that).

yaxu:
I'll add the library as an additional commit. Thanks for this!

xmacex:
Thank you @yaxu.

> Do you think the -f is needed?

My source for the need of `-f` on new Cabal is the GHC/Cabal thread [*ghc-pkg on 8.8.1 with cabal 3.0 does not list user installed packages*](https://gitlab.haskell.org/ghc/ghc/issues/17341), and solving it for myself, in a GHC 8.8.1 and Cabal 3.* environment.

Keep on raving  🔈🔈 we people on the internet really appreciate what you do!

yaxu:
Thanks @xmacex and other internet people, yes I needed -f too under Linux. This v2- stuff is a bit of a mess in my opinion.

PR #586, opened 2019-12-07
vol and clip

dktr0:


src/Sound/Tidal/Params.hs
```diff
@@ -179,6 +179,18 @@ end = pF "end"
 gain :: Pattern Double -> ControlPattern
 gain = pF "gain"
 
+-- | 'vol' is a version of gain with a built-in limiting that clips the provided
+-- values to stay between 0 and 1.999 in order to protect against common
+-- typing errors causing dangerously or uncomfortably loud sounds.
+vol :: Pattern Double -> ControlPattern
+vol = gain . clip 0 1.999
```
yaxu: Why not implement this in the target (e.g. scdirt/webdirt) rather than in tidal?
@telephon do you have a view on this?

src/Sound/Tidal/Params.hs
```diff
@@ -179,6 +179,18 @@ end = pF "end"
...
+
+-- | clip constrains a fractional pattern to stay within the limits
+-- specified by lowerBound and upperBound.
+clip :: Ord a => Pattern a -> Pattern a -> Pattern a -> Pattern a
+clip lowerBound upperBound x = f <$> lowerBound <*> upperBound <*> x
```
yaxu: I think it's worth making a version `_clip` for internal use (such as above) that doesn't have the overhead of patterned parameters. Then `clip` can be defined relative to that.
(Also I think it's best to use `*>` rather than `<*>` in this kind of situation, so that the structure of the final parameter takes precedence)

src/Sound/Tidal/Params.hs
dktr0: One could just as well ask why not here? Either way you have to change something in Tidal because both cases - safe and unsafe - need to be available to people who are live coding.

src/Sound/Tidal/Params.hs
dktr0: Not sure when I'll have time for this, so please feel free to just do what you suggest.

src/Sound/Tidal/Params.hs
dktr0: It's a choice between changing it in one place versus having to change it in three places. But I don't care too much how it gets implemented (we're talking about real small change of efficiency anyway) so long as the result is something safer for everyone.

yaxu:
vol / clip now not needed, gain is now assumed to be limited to safe level (e.g. 2) in the synth, with the non-limited 'overgain' added on top for full control. @dktr0 can you make a separate PR with the tidal-parse changes please, ta!

dktr0:
clip is still generally useful though

dktr0:
It would be less work if you just accepted this request and then removed the undesired vol - thanks!

yaxu:
Clip still needs some work as above

dktr0:
Please feel free to reimplement it however you like or remove - my attention is on other things.

yaxu:
OK, I'll spend some time deleting these for you, let's try to keep PRs cleaner in the future though.

dktr0:
I have no idea what you mean by cleaner in this case or the "for you" in your comment.

yaxu:
I guess by cleaner I mean not mixing topics in one PR, and updating the title to represent what's inside if things change
In the process of working on this issue I've done a fair amount of work and had a to-and-fro with @telephon getting my submissions to superdirt releasable despite not knowing much about its architecture, it's been a useful learning experience.
So I think it's good for PR submitters to take care of getting their code ready for release through feedback from other maintainers, in the process having knowledge passed on about how things work
Maybe we should have a discussion somewhere about social aspects of tidal development to keep this a happy experience for everyone though

dktr0:
This is a lot of conversation for a couple of lines of running code with zero consequences for users who don't explicitly choose to invoke them...  (I'm referring specifically to the interaction around this rather trivial pull request, not the somewhat larger issue of safety and limits in Tidal.)

I am engaged with "tidal development" very much as a limited tangent from other projects - (1) Estuary and (2) teaching live coding generally in classes, workshops, and ensembles. Well, admittedly I do find work on MiniTidal an enjoyable challenge as well. The "controversial" parts of this pull request were simply to share the results of an immediate fix (including the tangentially produced but possibly more useful "clip") with others.

In this context, if someone asks me to resubmit a PR because of an actual identified problem with it I would be happy to oblige.  I'm afraid I can't really afford the time to resubmit things because people would prefer they be implemented another way or because other unrelated and equally trivial things got dragged into the same pull request. The latter is just not really a problem at all in my view, and the former is just something where I think the onus is on the suggester to provide their alternate implementation if they feel strongly about it. If there's a discussion about social aspects of Tidal development I am sure I will follow that with interest, although I doubt it will move my own boundaries much in this respect.

I am now conscious that the time discussing this now far exceeds the time it would take either of us to delete vol and reimplement clip in the "official" pattern (although I am sure you would do it faster...)...

yaxu:
The problem was that vol wasn't needed and clip wasn't right. I already removed both. I don't think implementing clip is a priority, as min and max work fine.

PR #595, opened 2019-12-28
add newline so atom package can handle BootTidal.hs correctly

bgold-cosmos:


ndr-brt:
could be cleaner with the `:{ :}` block? e.g.
```haskell
:{
import Sound.Tidal.Context
import System.IO (hSetEncoding, stdout, utf8)
hSetEncoding stdout utf8
:}
```

bgold-cosmos:
I don't think multiple imports within a `:{` `:}` block work in the current version of GHCi (I know it was broken in the past).

yaxu:
Thanks will release this in the morning.. 
It works for me in emacs, which I think loads the file using :script
Maybe atom should do the same?

bgold-cosmos:
Yes, long-term that's probably better, but right now it uses the same reformatting functions that ctrl-enter uses.

ndr-brt:
Ok, new issue:
https://github.com/tidalcycles/atom-tidalcycles/issues/85

yaxu:
I've released it now, thanks for the catch @bgold-cosmos

Issue #579, opened 2019-12-02
scales only support integers


quakehead:
I'd like to be able to create Arabic scales ([maqams](http://maqamworld.com/en/maqam.php)) many of which use quartertone notes.

The scales here use the typeclass Num, which make it seem like I could use a fraction, but it's Haskell error.
https://github.com/tidalcycles/Tidal/blob/master/src/Sound/Tidal/Scales.hs

When I try to add
```
bayati :: Num a => [a]
bayati = [0,0.5,4,6,8,8.5,12]
```
I get this error.
```
   • Could not deduce (Fractional a) arising from the literal ‘0.5’
      from the context: Num a
        bound by the type signature for:
                   bayati :: forall a. Num a => [a]
        at :5890:1-22
      Possible fix:
        add (Fractional a) to the context of
          the type signature for:
            bayati :: forall a. Num a => [a]
    • In the expression: 0.5
      In the expression: [0, 0.5, 4, 6, ....]
      In an equation for ‘bayati’: bayati = [0, 0.5, 4, ....]
```

@yaxu mentioned on the chat that it could be a simple case of a type change. For example, this is valid Haskell:
```
bayati :: Fractional a => [a]
bayati = [0,0.5,4,6,8,8.5,12]
```

**Environment (please complete the following information):**
 - Haskell version: The Glorious Glasgow Haskell Compilation System, version 8.6.5
 - Tidal version 1.4.4

**Additional context**
 http://maqamworld.com/en/maqam/bayati.php

quakehead:
I changed the typeclass of scales to Fractional and it compiled.

Though the frequencies that actually gets sent to SuperCollider (or SuperDirt) are off - looks like the quarternotes are rounded. I used SuperCollider IDE's Server OSC dump to see the frequencies.

I used this to play notes in a scale
`d1 $ (scale "bayati" "0..10") # s "superpiano" # sustain "1"`


Bayati [0,0.5,4,6,8,8.5,10]

261.626 0
261.626 0.5
329.628 4
369.994 6
415.305 8
261.626 8.5
466.164 10
523.251 12 Octave
523.251 12.5
659.255 16


The pattern seems ok:
```
(scale "bayati" "0..8") :: Pattern Double
t>(0>⅑)|0.0
(⅑>²₉)|0.5
(²₉>⅓)|4.0
(⅓>⁴₉)|6.0
(⁴₉>⁵₉)|8.0
(⁵₉>⅔)|8.5
[...]

PR #596, opened 2019-12-28
Support fractional scales and add Arabic scales

quakehead:
I made scales Fractional instead of Num so I could add 24 note scales (or quartertones), fixing #579

The scale has to be used with `note` or `n` to work with fractions

`d1 $ note (scale "bayati" "0 .. 8") # sound "superfork"`

I added 6 new scales.

There's probably a more elegant way to refactor this, to parametrize the number of notes in a scale (mainly 12 or 24) like how it's done in supercollider:
https://github.com/supercollider/supercollider/blob/f65a1ef7bfb100f4ccb7dcab3a6dff6826be4344/SCClassLibrary/Common/Collections/Scale.sc

### DIFF
--- a/test/Sound/Tidal/ScalesTest.hs
+++ b/test/Sound/Tidal/ScalesTest.hs
@@ -19,249 +19,249 @@ run =
             it "can transform notes correctly over 2 octaves - minPent" $ do
                 compareP (Arc 0 1)
                     (Sound.Tidal.Scales.scale "minPent" twoOctavesOf5NoteScale)
-                    ("0 3 5 7 10 12 15 17 19 22"::Pattern Int)
+                    ("0 3 5 7 10 12 15 17 19 22"::Pattern Rational)
             it "can transform notes correctly over 2 octaves - majPent" $ do
                 compareP (Arc 0 1)
                     (Sound.Tidal.Scales.scale "majPent" twoOctavesOf5NoteScale)
-                    ("0 2 4 7 9 12 14 16 19 21"::Pattern Int)
+                    ("0 2 4 7 9 12 14 16 19 21"::Pattern Rational)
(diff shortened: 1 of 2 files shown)


test/Sound/Tidal/ScalesTest.hs
```diff
@@ -270,49 +270,51 @@ run =
...
                     (Sound.Tidal.Scales.scale "messiaen6" twoOctavesOf8NoteScale)
-                    ("0 2 4 5 6 8 10 11 12 14 16 17 18 20 22 23"::Pattern Int)
+                    ("0 2 4 5 6 8 10 11 12 14 16 17 18 20 22 23"::Pattern Rational)
             it "can transform notes correctly over 2 octaves - messiaen7" $ do
                 -- semitone, semitone, semitone, tone, semitone, semitone, semitone, semitone, tone, semitone
                 compareP (Arc 0 1)
                     (Sound.Tidal.Scales.scale "messiaen7" twoOctavesOf10NoteScale)
-                    ("0 1 2 3 5 6 7 8 9 11 12 13 14 15 17 18 19 20 21 23"::Pattern Int)
+                    ("0 1 2 3 5 6 7 8 9 11 12 13 14 15 17 18 19 20 21 23"::Pattern Rational)
+-- {-
```
quakehead: oops, should fix this

quakehead:
No rush about this but any comments? Should I rewrite this in terms of degrees and to keep scales as integers?

yaxu:
Sorry I missed this! Thanks for the nudge. So I can fully understand the problem this solves, can you share a quick code example that didn't work before this change please?

yaxu:
OK I think I see, Num is too general. I'll merge it. Thanks!

PR #598, opened 2020-01-09
overgain and overshape in tidal-parse which is updated to work with Haskellish-0.2.0

dktr0:


yaxu:
I think the travis errors are just down to new haskellish not being in hackage yet

dktr0:
Ah right I forgot to do that! (An easy mistake to make when working on Estuary and Punctual because we pull specifically pinned packages in nix so the hackage part doesn't play much of a role...). Will do today.

yaxu:
No problem! 
I don't suppose you mind explaining what nix is in this context, and how it relates to cabal new- stuff?

dktr0:
Haskellish-0.2.0 is in hackage now. nix provides a way of specifying how software is built where all of the dependencies (and I really mean *all* - it includes build and OS tools) are pinned and sandboxed. This provides a quite strong guarantee of reproducible builds (and buildability) at the cost of sometimes building more than might strictly speaking be necessary. When its used with Haskell cabal is still an important part, although my experience has been that I am somewhat shielded from the vagaries of it. We're using cabal new-build when we do incremental, development builds of Estuary, and I believe it is basically cabal new-build when we ask nix to do the full release build as well. It's a bit of a pain to maintain the nix configuration file although it is totally worth it for avoiding the absolute dependency hell we would have in a project as big as Estuary otherwise...

PR #616, opened 2020-03-24
typo? a sixteenth is 0.0625

mxmxyz:


yaxu:
Thanks!

yaxu:
'f' should be '0.2' not '0.25' as well, will fix that..

PR #617, opened 2020-03-29
Abridge accumulate function using scanl

benjwadams:


yaxu:
That's tidier, thanks!

PR #630, opened 2020-04-17
export drawLineSz

mxmxyz:
Personally I think making `drawLineSz` usable, rather than just keeping it internal, would be nice? It allows one to do cool things such as screen-flooding live, and to visualize very long/complex patterns. I don't really see why not?


yaxu:
Agreed!

PR #640, opened 2020-04-27
new random implementation

dktr0:
based on xorshift algorithms, this seems to be faster to compute than the previous model, and probably less periodic. The same implementation (with slightly different naming conventions) has also been committed to the tempi library where it is available for other languages/projects.


yaxu:
Thanks @dktr0 ! Looks like a strange lockup on one of the travis builds, hopefully just intermittent..

Issue #615, opened 2020-03-23
tidal.el on Linux expects Tidal to be installed through Cabal


pashultz:
Arch Linux has [a package for TidalCycles](https://www.archlinux.org/packages/community/x86_64/tidalcycles/), which installs BootTidal.hs under /usr/share/x86_64-linux-ghc-8.8.3/tidal-1.4.8/ (version numbers and architectures may change, of course). Unfortunately the code in tidal.el has a path hardcoded under ~/.cabal/store/, as of 
https://github.com/xmacex/Tidal/commit/59fe608d77cfe9ea33504e8f6a0c9b35706c63e6 (Note that the message on that commit anticipates this problem: "what if libraries are not under user's home directory?") This means tidal.el can't find Tidal when it's installed systemwide.

**To Reproduce**
Steps to reproduce the behavior are mostly the same as in https://github.com/tidalcycles/Tidal/issues/583:
1. Install `emacs` and `tidalcycles` through the Arch Linux package system, along with SuperDirt and any other requirements.
2. Open a .tidal file (C-x C-f my-humps.tidal), and verify that tidal-mode is active.
3. Run the `tidal-start-haskell` command with C-c C-s

**Expected behavior**
There should be a `tidal>` prompt in the Comint buffer, but it fails with "file not found" messages. 

Here's the offending passage:
```
'(("path" . "ghc-pkg field -f ~/.cabal/store/ghc-$(ghc --numeric-version)/package.db tidal data-dir")
             ("separator" . "/")
```
If I delete the hardcoded path (removing the -f option), it loads just fine:
```
'(("path" . "ghc-pkg field tidal data-dir")
             ("separator" . "/")
```
I assume the -f option was added because ghc-pkg [doesn't otherwise know](https://gitlab.haskell.org/ghc/ghc/issues/17341) about the user's .cabal. Is there a fallback mechanism that would let it check .cabal first, then fall back to the systemwide database if that fails?

**Environment:**
 - OS: Arch Linux, x86_64
 - Haskell version: 8.8.3
 - Tidal version: 1.4.8-13 (as packaged [here](https://www.archlinux.org/packages/community/x86_64/tidalcycles/))
 - SuperCollider version: 3.11.0-beta1

Thanks in advance for your help!

pashultz:
[...]

PR #652, opened 2020-05-17
Simplify tidal-boot-script-path in tidal.el

jamagin:
Newer layouts of ~/.cabal appear to structure things differently,
and as suggested in tidalcycles/Tidal#615 this works for the
Unix case.


yaxu:
Thanks!

PR #664, opened 2020-06-04
fixes #662

### LINKED ISSUE #662: repair instance NFData (Pattern a)
I doubt that the behaviour of
```
instance NFData a =>  NFData (Pattern a) where 
    rnf (Pattern q) = rnf $ \s -> q s
```
was intended. It should be ` .. = rnf q`? Then it could be derived generically (see #661 )

See discussion at https://github.com/tidalcycles/Tidal/issues/606#issuecomment-638400358

jwaldmann:
Contains the trivial fix. 

I added a test case but I did not find `shouldThrow` in Microspec, so it looks messy.


yaxu:
Thanks!

PR #665, opened 2020-06-04
Fix 661

jwaldmann:
This is on top of PR for #662:
* it keeps the test case
* it replaces instance NFData (Pattern a), with the generic one


yaxu:
Hm so this is a includes a commit I've already merged from - https://github.com/tidalcycles/Tidal/pull/664

Was this intended? Should I have only merged this one?

jwaldmann:
It's fine. There may be more changes around NFData/strictness coming  because #667 anyway.

Issue #439, opened 2018-12-24
How to use `round` with `n`?


jarmitage:
Sometimes you want to generate a melody using a continuous function, for example:

`n (segment 12 $ range 0 12 $ sine)`

Rather than generating the numbers 0-12, this generates:

```
(0> 1/12)|6.0
( 1/12>⅙)|9.0
(⅙>¼)|11.196152422706632
(¼>⅓)|12.0
(⅓> 5/12)|11.196152422706632
( 5/12>½)|9.0
(½> 7/12)|6.000000000000002
( 7/12>⅔)|2.999999999999999
(⅔>¾)|0.8038475772933698
(¾>⅚)|0.0
(⅚> 11/12)|0.8038475772933684
( 11/12>1)|2.9999999999999973
```

If you wanted this melody to be clamped to semitones you would previously do this:

`n (round <$> segment 12 $ range 0 12 $ sine)`

Which means you can then do stuff like this:

`n (scale "major" $ round <$> segment 12 $ range 0 12 $ sine)`

However, now that `n` is a Pattern Double, this is no longer possible, because `round` causes an exception:

```
tidal> *** Exception: round: not supported for patterns
CallStack (from HasCallStack): error, called at src/Sound/Tidal/Pattern.hs:392:13 in tidal-1.0.4:Sound.Tidal.Pattern
```

For the same reason it's also not possible to do:

`n (fromIntegral $ round <$> segment 12 $ range 0 12 $ sine)`

Maybe there's a simple Haskell-y answer to this, but I thought it was worth noting since you could describe this as an unintended consequence of moving `n` to doubles.

I guess what would be ideal is something similar to this:

`n (round <$> segment 12 $ range 0 12 $ sine)`

But produces output similar to:

`n "0 1 2 3 4 5 6 7 8 9 10 11 12"`

i.e. 

```
(0> 1/13)|n: 0.0f
( 1/13> 2/13)|n: 1.0f
( 2/13> 3/13)|n: 2.0f
( 3/13> 4/13)|n: 3.0f
( 4/13> 5/13)|n: 4.0f
( 5/13> 6/13)|n: 5.0f
( 6/13> 7/13)|n: 6.0f
( 7/13> 8/13)|n: 7.0f
( 8/13> 9/13)|n: 8.0f
( 9/13> 10/13)|n: 9.0f
( 10/13> 11/13)|n: 10.0f
( 11/13> 12/13)|n: 11.0f
( 12/13>1)|n: 12.0f
[...]

PR #673, opened 2020-06-07
Updates to src/Sound.Tidal.UI.hs,  change quantise to round, add qfloor and qceiling ref #439

lwlsn:
Issue # 439


yaxu:
Thanks!

Issue #614, opened 2020-03-18
Make the first parameter of chunk a pattern


yaxu:

yaxu:
Thanks @lwlsn!

PR #685, opened 2020-06-17
Updates to chunk function to accept Pattern Int, give chunk some test patterns, from issue #614

lwlsn:


yaxu:
Looks like the build failures are related to tidal-parse, will merge and then fix

yaxu:
(thanks!)

PR #687, opened 2020-06-20
tidal-boot-script-path back again for Darwin

yasushisakai:
commit da08b865316afaa0b6b18fc4be76f64b1b62b5c2 simplifies the path for
boot script path, but latest installation in mac os (10.15.5) works with the
previous configuration.

```
$ ghc-pkg field tidal data-dir
ghc-pkg: cannot find package tidal
$ ghc-pkg field -f ~/.cabal/store/ghc-$(ghc --numeric-version)/package.db tidal data-dir
data-dir: /Users/yasushi/.cabal/store/ghc-8.8.3/tdl-1.6.0-6feaeacf/share
```

this commit separates the clause between `darwin`(macos) and `gnu/linux`

relates to #615 specifically https://github.com/tidalcycles/Tidal/issues/615#issuecomment-638128702

```
% ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.8.3
% cabal --version
cabal-install version 3.2.0.0
compiled using version 3.2.0.0 of the Cabal library 
```


yaxu:
Ah, thanks! I feel that this is probably not a difference between linux and mac, but between different ways of installing tidal.

My guess is that `ghc-pkg field tidal data-dir` only works if you installed tidal with `cabal v1-install tidal`, and `ghc-pkg field -f ~/.cabal/store/ghc-$(ghc --numeric-version)/package.db tidal data-dir` only works if you installed it with `cabal install tidal --lib`. Does that fit with your experience?

If so we should open an issue upstream on cabal, so `ghc-pkg field tidal data-dir` works in all cases.

yasushisakai:
> Does that fit with your experience?

yup I suppose that's the case!

I'm not as fluent in elisp nor haskell and a total newbie around `cabal`. and yes, I've installed tidal through `cabal install tidal --lib` as from the userbase docs. so feel free to throw this PR away.  

I wonder how the other editor environments are handling this? (vscode and atom worked for me to)

yaxu:
I think it uses the same command, but tries harder to find the path to ghc-pkg, and also has its own copy of BootTidal.hs as a fallback: https://github.com/tidalcycles/atom-tidalcycles/blob/master/lib/ghc.js

tonyday567:
#751 fixed this.

yaxu:
Thanks all

PR #689, opened 2020-06-23
Patternize the first parameter of fit

bgold-cosmos:


yaxu:
Thanks! I'll fix up tidal-parse

BenjaminWynn:
Thanks so much for this!!

And pardon my ignorance on how GitHub works, but since the change has been committed does that mean that I can update using the normal method?
```
cabal v1-update
cabal v1-install tidal
```

jwaldmann:
`cabal install` pulls from hackage https://hackage.haskell.org/package/tidal

Any change here in the repo needs to be pushed there (by Alex). He normally collects several changes before doing so.

yaxu:
I've just done a release now so that should work.

BenjaminWynn:
Got it. Thanks so much all!

PR #691, opened 2020-07-05
MIDI Notes from GM2 Percussion key map spec.

lvm:
Adds missing midinotes for drum synths.
In the past i maintained this https://github.com/lvm/tidal-midi-gm but I think this could be useful in the core.

Ref:
https://www.voidaudio.net/percussion.html
https://www.musicrepo.com/wp-content/uploads/GM2-Drum-Map.txt


yaxu:
Nice, impressive you got 2 char aliases for them all!

lvm:
Hey! thanks. Yeah! let's see how this works out. Probably 90% of the people (myself included) will still use the same 16 midi notes with the mapping table on the side just in case :-)

lvm:
@yaxu 
i just noticed that labeled "ss" (Side stick) when it should be "rm" (Rim Shot). Do you prefer to replace this yourself or want me to make a new PR? 
This is the line: https://github.com/lvm/Tidal/blob/main/src/Sound/Tidal/Params.hs#L583

PR #692, opened 2020-07-06
Replace "ss"  for a more standard "rm"

lvm:
This would also match the sample directory.


yaxu:
Thanks! I guess you could have both standing for the same number, if you wanted.

lvm:
Thanks again.
Probably I'll create a new PR soon with some variants. I've been reading this and there are three types: 
* Orchestral
* Marching band
* General (the one that was updated)

https://usermanuals.finalemusic.com/SongWriter2012Win/Content/PercussionMaps.htm

PR #696, opened 2020-07-10
add qualified imports

bgold-cosmos:
BootTidal.hs needs to make sure some modules are imported for `getcps` to work properly, and I also took the opportunity to provide a `getnow` command that works similar to a function that existed before the reworking of Stream.hs


yaxu:
Great, thanks a lot!

Issue #698, opened 2020-07-21
make irand patternable


ndr-brt:
would be nice to have irand patternable.

ndr-brt:
Closed via #701

PR #701, opened 2020-07-23
Irand patternable

ndr-brt:
As described in #698 
Is my first "haskell" PR, please be kind with my "beginnerness" :)

I created a test to demonstrate the functionality, and also add some test to cover `irand` (that wasn't covered)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,10 @@ cabal
 .stack-work
 .DS_Store
 .ghc.environment.*
+out/
+.idea/
+*.lock
+*.iml
 
 # text editor temp files
 *~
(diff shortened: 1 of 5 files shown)


yaxu:
Thanks! I'm a little suspicious of the use of `>>=` here, which takes structure from both the first argument and the second 'inner' one. Normally I'd use `innerJoin` for parameters, to take structure from the final parameter. But judging by your tests the `>>=` seems to work well, with the inner pattern being a continuous one. Interesting!

ndr-brt:
I wanna be honest... I took some "inspiration" from the `run` implementation (https://github.com/tidalcycles/Tidal/blob/04b889742fca9923ee5bc871572b34db6e6010a6/src/Sound/Tidal/Core.hs#L195), then the tests gave me confidence :)

Issue #683, opened 2020-06-15
Rename master to main


yaxu:
We should rename the 'master' branch to 'main'. Trying this now, opening this in case of unintended issues..

jwaldmann:
What's the goal here? Do you want to use "master" for something else?

kindohm:
it's a metaphor that we should not use any more: https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.1

yaxu:
Yes github are switching to 'main' as default branch name but I guess won't do this change to older projects.

I removed master/slave metaphors from the source in tidal-1.0

yaxu:
@brandondyer64 why the downvote, is this causing you problems?

BrandonDyer64:
I believe that changing it makes things inconsistent.

I also believe that the terminology draft is referring to master-slave (as stated in the title of the section)

"master" in git does not mean master of slaves, but rather original copy.

I also don't think it should be changed until GitHub have actually made their change.

I'm not here to tell you how to run your repository. The -1 was just to show my disagreement with the change. I'm not going to complain if you do.

yaxu:
@BrandonDyer64 Do you actually use Tidal or are you just a random passing white man with an opinion?

froozeapp:
There's also no "slave" in git what makes this even funnier. Also, should we ban the entire BDSM community if the terminology is so bad? Kind of off-topic, I know

yaxu:
Dear trolls, if you're spending extended amounts of time scanning github issues, in order to tell people not to waste time on a minute change, then maybe you should reconsider your priorities.

bgold-cosmos:
Alex, thanks for taking the time to deal with this.  IIRC, before git were SVN/CVS where it was called "trunk" anyway.  I don't understand why some people get so worked up when _other_ people change a name in their own code.

dktr0:
Yes, thanks Alex. Going to make similar changes in a bunch of my repositories now too.

froozeapp:
[...]

PR #703, opened 2020-07-24
Set travis badge on the right branch

ndr-brt:
It was by default on `master`, that does not exists anymore due to #683


yaxu:
Ah I was wondering about this.. Thanks!

PR #705, opened 2020-08-06
Added grain custom parameter

khoparzi:
Remaps begin and end to begin and width, to create a granulising effect that can be patterned
eg. 
`d1 $ s "bev" >| grain ((1/16) * "0 3 4 6*<1 2 3>") (1/64)`


yaxu:
interesting, thanks!

yaxu:
I backed this out as it didn't compile.. I think `grain` would be better in `Sound.Tidal.UI` so that it has access to `#`
It's worth running `cabal v1-test` before committing. Also please add type signatures for functions, thanks!

Issue #702, opened 2020-07-24
Ratio alias "e" evaluated not correctly on delay


ndr-brt:
I noticed this strange behaviour with ratio aliases and delay.

This pattern:
```haskell
d1 $ s "bd" $ delay "h"
```
works correctly, as in the osc message the parameter ```delay``` is 0.5
but, using `e`:
```haskell
d1 $ s "bd" $ delay "e"
```
in the osc message the parameter ```delay``` is 4

probably it's interpreted as the note `e` (the fifth semitone, so, number 4)

ndr-brt:
I noticed this problem also with all the other parameters

ndr-brt:
It works correctly used for temporal alias like:
```
d1 $ off "e" (|* speed 2) $ s "bd"
```

yaxu:
Yes it works for temporal and other rational numbers. For floating point numbers, musical notes do indeed get in the way. 

One way around this could be to make `n` and `note` a different type, so that they are parsed in a different way. This would likely cause complications, though.

ndr-brt:
And what about defining those parameters (at least those that do not concern notes) as rational parameters?
I tried with a unit test, defining a rational parameter:
```haskell
pR :: String -> Pattern Rational -> ControlPattern
pR name = fmap (Map.singleton name . VR)
```
then setting the parameters as `pR` instead of `pF`, it works correctly.
For a parameter like "delay" but also... "gain" or any other parameter "0 to 1" I think it works better.
Do you see any possible issue with that?

yaxu:
The only problem is efficiency - as far as I know, floats are much more efficient than rationals.. and ultimately it'll be a float in the resulting OSC message. I think it's better to create different 'flavours' of float values.. So as well as VF float values we could have VN for notes, the only difference being in how the parsing is handled.

ndr-brt:
Ok, I writed a pair of unit test and played a little, but I eventually need some help...
I introduced a
```haskell
newtype Note = Note Double
```
and duplicate some code to make it work associated at the value `VN`
I have some problems implementing a `pDouble` that works with `Note`.

Any suggestion?

jwaldmann:
+1  I can look into the newtype stuff if you share your code (and highlight the problem)

ndr-brt:
created PR #708 , when you can take a look :)
thanks for the help
@jwaldmann @yaxu

jwaldmann:
OK I am looking at it now. 

The type error goes away after
[...]

PR #708, opened 2020-08-26
Introduce Note newtype

ndr-brt:
Sorry in advance for this PR with failing build.
As I said in #702 had some issues trying to make the Note newtype work.

There are some test cases in `ParamsTest.hs`, the compilation error is on the `pNote` function definition, and I leaved some commented out code in the "Data compare" section.

Just need some help :)

### DIFF
--- a/src/Sound/Tidal/Pattern.hs
+++ b/src/Sound/Tidal/Pattern.hs
@@ -353,8 +353,8 @@ instance Ord Value where
   compare (VN x _) (VI y _) = compare x (fromIntegral y)
   compare (VI x _) (VN y _) = compare (fromIntegral x) y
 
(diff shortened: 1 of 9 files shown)


src/Sound/Tidal/Pattern.hs
```diff
@@ -47,8 +47,8 @@ type Time = Rational
 newtype Note = Note Double deriving (Typeable, Data, Generic, Eq, Ord, Show, Enum, Num, Fractional)
```
jwaldmann, line 47: you could write `newtype Note = Note { unNote :: Double }`.

ndr-brt: done :)

yaxu:
Looks great, thanks a lot!

PR #720, opened 2020-09-27
added to tidal-parse: toScale, arpeggiate, arp, weave, weaveWith, ghost, inside, outside, inv

dktr0:
I'm not sure why there are a bunch of old commits being included in this pull request - I thought most of that was merged ages ago.


yaxu:
I think it's probably harmless but have a look at contributing.md for an approach to merging upstream stuff that might help avoid the commit confusion

yaxu:
It looks like the Travis build is more generally broken, for now it's worth making sure cabal test runs OK locally if you haven't already

dktr0:
thanks - yeah, I did cabal test before committing.

dktr0:
Not sure I follow the comment about contributing.md. Are you suggesting the confusion happens because my branch is also named main?

yaxu:
@dktr0 Sorry was on my phone. 

I meant this bit: https://github.com/tidalcycles/Tidal/blob/main/CONTRIBUTING.md#recommendations-to-handle-forks-and-branches

When I update my fork by creating a PR from upstream to downstream, I end up with the commits reappearing when I try making a PR back to upstream. The procedure described there seems to avoid this. YMMV!

PR #722, opened 2020-09-30
Reimplementing grain custom parameter

khoparzi:
Remaps begin and end to begin and width, to create a granulising effect that can be patterned
eg.

d1 $ s "bev" >| grain ((1/16) * "0 3 4 6*<1 2 3>") (1/64)

or for a simpler grain selection with

d1 $ grain' "0.0125:0.06125" # s "bev"

### DIFF
--- a/src/Sound/Tidal/Params.hs
+++ b/src/Sound/Tidal/Params.hs
@@ -207,6 +207,9 @@ gain = pF "gain"
 gate :: Pattern Double -> ControlPattern
 gate = pF "gate"
 
+grain' :: Pattern String -> ControlPattern
+grain' = grp [mF "begin", mF "end"]
+
 hatgrain :: Pattern Double -> ControlPattern
 hatgrain = pF "hatgrain"
 -- | a pattern of numbers from 0 to 1. Applies the cutoff frequency of the high-pass filter.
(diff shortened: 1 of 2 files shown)


yaxu:
Do you think this is needed as well as the `slice` and `splice` functions @khoparzi ? I think these are probably the same (haven't tested):

```
d1 $ s "bev" >| grain ((1/16) * "0 3 4 6*<1 2 3>") (1/64)

d1 $ slice 64 (4 * "0 3 4 6*<1 2 3>") $ s "bev"
```

khoparzi:
Perhaps the example I posted is very similar to the slice I splice functions. I personally use the grain parameter more as a way of getting a microsound type effect that is not easily divided into even parts specially for vocal samples/breaks that are not chopped with finesse while collecting.

```
d1 $ stack [
  fast 2 $ grain 0.4 0.04, -- Just plays the synth note in bev
  fast 4 $ grain 0.2 0.04, -- Just plays a selection of two phonemes
  segment 8 $ grain (slow 32 $ sine) 0.012 -- This line essentially emulates palindrome $ striate to me but with finer control
] # s "bev"
```

Or patterning the width of the selection with

`d1 $ s "foo*4" # grain 0 "0.1 0.2 0.3"`

khoparzi:
This video explains shows off some uses of the width parameter in performance  https://youtu.be/La-JcuoaxZE

yaxu:
Thanks for clarifying!

PR #726, opened 2020-10-02
Handle Note type in tidal-parse

### LINKED ISSUE #725: Broken tests in tidal-parse
After merge of #708, tests on tidal-parse are failing, caused by introduction of the new `Note` datatype

ndr-brt:
Should fix #725 

Test are still broken but with an error like
`*** Failed! Exception: '<<loop>>' (after 1 test): | Exception thrown while showing test case: '<<loop>>' |`

Think it's something not bind to the `Note` newtype. 
@dktr0 can you help me about that?

### DIFF
--- a/tidal-parse/src/Sound/Tidal/Parse.hs
+++ b/tidal-parse/src/Sound/Tidal/Parse.hs
@@ -710,7 +710,7 @@ instance Parse (Int -> Pattern Double -> Pattern a -> Pattern a) where
   parser = $(fromTidal "degradeOverBy")
 
 instance Parse (Int -> Pattern T.Note -> Pattern a -> Pattern a) where
-  parser = parser
+  parser = empty
 
 instance Parse ((a -> b -> Pattern c) -> [a] -> b -> Pattern c) where
   parser =
@@ -751,7 +751,7 @@ instance Parse (Pattern Double -> (Pattern a -> Pattern a) -> Pattern a -> Patte
     $(fromTidal "plyWith")
 
 instance Parse (Pattern T.Note -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a) where
-  parser = parser
+  parser = empty
 
 instance Parse ([Int] -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a) where
   parser = $(fromTidal "foldEvery")
--- a/tidal-parse/src/Sound/Tidal/Parse.hs
+++ b/tidal-parse/src/Sound/Tidal/Parse.hs
@@ -6,8 +6,6 @@ import           Language.Haskell.Exts
 import           Language.Haskellish as Haskellish
 import           Control.Applicative
 import           Data.Bifunctor
-import           Control.Monad
-import           Data.Either
 import           Control.Monad.Except
 
 import           Sound.Tidal.Context (Pattern,ControlMap,ControlPattern,Enumerable,Time)
@@ -23,12 +21,6 @@ parseTidal = f . parseExp
     f (ParseOk x) = second fst $ runHaskellish parser () x
     f (ParseFailed _ "Parse error: EOF") = return T.silence
     f (ParseFailed l s) = throwError $ show l ++ ": " ++ show s
-
-{- test :: Parse a => String -> Either String a
-test = f. parseExp
-  where
-    f (ParseOk x) = value x
-    f (ParseFailed l s) = Left $ show l ++ ": " ++ show s -}
 
 -- The class Parse is a class for all of the types that we know how to parse.
 -- For each type, we provide all the ways we can think of producing that type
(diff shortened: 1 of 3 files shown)


tidal-parse/src/Sound/Tidal/Parse.hs
```diff
@@ -708,6 +750,9 @@ instance Parse (Pattern Double -> (Pattern a -> Pattern a) -> Pattern a -> Patte
     $(fromTidal "someCyclesBy") <|>
     $(fromTidal "plyWith")
 
+instance Parse (Pattern T.Note -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a) where
+  parser = parser
```
jwaldmann, line 754: This looks strange, and may create the `<<loop>>` when called? Some `fmap T.Note` missing?

dktr0: Yeah, that's definitely infinite/empty recursion. If you have, for example an example of a function with that type you could do: parser = $(fromTidal "thatFunction")

ndr-brt: uhm... ok... and what if I haven't? at the moment only params "note", "midinote" and "up".
I tried also with `empty`, but, no luck

dktr0: If there's no way of making that type (Pattern Note -> (Pattern a -> Pattern a) -> Pattern a) in Tidal then you don't need to make a parser/Parse instance for it at all. I can comment more if you can give some examples of actual code you might want to parse.

jwaldmann: What exactly *is* the failing test here? (Is it visible in Travis, or how do I reproduce?)

ndr-brt: I have to tell you... I didn't need to parse anything, I was just trying to fix compilation errors and failing tests, but I'm not fully aware of the behavior of tidal-parse code, so any help would be appreciated :)

ndr-brt: @jwaldmann here's the errors, at the end of the log:
https://travis-ci.org/github/tidalcycles/Tidal/jobs/732245834

dktr0: I will look at this later this afternoon. Still it would help to have some examples of this Note stuff in action - being mentioned here is the first I've heard of any of this...

ndr-brt: Note is just a sort of alias of Double, born to solve this issue: https://github.com/tidalcycles/Tidal/issues/702

dktr0: I see two places where you have parser = parser, which creates an infinite loop. If there is really no expression that evaluates to the types (at those places where you currently have parser = parser) then those instances/parsers (all of them) shouldn't be there at all; moreover, any pathways in other instances/parsers that reference those will also need to be deleted. I would expect it would "work" if both/all were defined as empty - perhaps you had defined one of them as empty and left the other in for the test?  Anyway, can I suggest that we start by removing pathways that don't correspond to any actual Tidal code - and then if it still causes problems, I'll look closer? (As the person usually maintaining this slightly complicated structure, I don't want to deal with the extra cognitive load of code that doesn't do anything!)

jwaldmann:
> If there is really no expression that evaluates to the types 

well the test cases are precisely such expressions? 

I am looking into this. For later reference, here's how I reproduce (with ghc-8.10.2)
```
cabal run --allow-newer tidal-parse test
```
I am getting exactly the output that's in the travis log linked above, ending
```
Successes: 57, Pending: 0, Failures: 9
```

dktr0:
@jwaldmann these are old tests that are failing - not tests that address expressions that don't exist... if there's no example of an expression for a given type then there's no parser that can be written for it, and no test of that parser that can be written. Meanwhile, parser = parser can certainly break existing tests.

dktr0:
If we say that one of the ways a can be produced is by parsing (b -> a) and then applying b, yet the parser we provide for (b -> a) is infinitely recursive and there is no way of actually producing (b -> a) that will cause many tests involving parsing a to fail.

ndr-brt:
@dktr0 woa, that was ees, I was wrong (and really silly) with `parser = parser`, changing that with `parser = empty` fixed everything.
Now we're ok!

dktr0:
Great! I still think code paths that are empty should be deleted. I'll probably do it instinctively later if you don't do it now, anyway... ;)

ndr-brt:
I did it by myself, removed all unused code parts, resolved also some compilation warnings. :)

dktr0:
Awesome - thanks!

dktr0 (approved):

yaxu:
Thanks a lot!

PR #727, opened 2020-10-04
Add missing functions pB and pR

thgrund:
I added the two functions pB and pR to create own pattern functions with the types Bool and Rational. 
The function for Pattern Time is still missing. This is because there is no function `VT`.


yaxu:
Thanks!

PR #728, opened 2020-10-07
Replace round with ceiling in qtrigger.

thgrund:
The function `qtrigger` does not work consistently because round maybe applies the offset for quantization to be in the past. If this happens, the trigger function will be ignored. The ceiling function always sets the offset to the beginning of the next cycle.

### DIFF
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -400,7 +400,7 @@ msec p = ((realToFrac . (/1000)) <$> cF 1 "_cps") *| p
 _trigger :: Show a => Bool -> a -> Pattern b -> Pattern b
 _trigger quant k pat = pat {query = q}
   where q st = query ((offset st) ~> pat) st
-        f | quant = (fromIntegral :: Int -> Rational) . round
+        f | quant = (fromIntegral :: Int -> Rational) . ceiling
           | otherwise = id
         offset st = fromMaybe (pure 0) $ do p <- Map.lookup ctrl (controls st)
                                             return $ ((f . fromMaybe 0 . getR) <$> p)


yaxu:
My thinking for using `round` for a quantised trigger was that if you have a pattern running, and want to retrigger it, you'll want to do that to the nearest cycle boundary.. otherwise if you miss the deadline by a tiny amount, it will wait until the next cycle. I can see that `ceiling` would be better in some circumstances though.. Not sure what the name of that should be.

thgrund:
Ah okay, that sounds logical.

The name ideas that come to my mind quickly could be:
- qtrigger'
- ntrigger : for next trigger 
- nctrigger : next cycle trigger
- ctrigger : ceiling trigger

Or should I think about it again? 
It would also be no problem for me to open a new PR that contains the round trigger and a ceiling trigger. For this I would continue to use this PR to keep the conversations in one place.

fatmilktv:
As spoken elsewhere, this is a hardware
software non-issue with great c++ bones. No 🤣🤣🤣 intra-darpa posting!!
This is an email list!! I love the ass of science!!!! Thus ends the prayer.

fatmilktv:
Suggest cherry pick unique.

jwaldmann:
An instance of https://news.ycombinator.com/item?id=24643894 ?

yaxu:
Still not sure what to suggest for this. Some of them look a bit too close to a racial slur for comfort when projected during a performance!

For quantising values there is the default `quantise` plus `qfloor`, `qceiling` and `qround`. So maybe it would make sense to have `trigger` with no quantising, and `tfloor`, `tceiling` and `tround` as aliases. I think these are a bit difficult to remember, like `tceiling` doesn't really say `trigger at the next cycle` for me. So perhaps there could be an alias `tnext = tceiling`. But how about `tround`? Maybe `tnearest`, but I think there could be a better name..

Or maybe this is all too fiddly and we should just change `qtrigger` to use round (as I think this is what people generally want) and add a `triggerWith` if people want to use other roundings.. That feels a bit cleaner

yaxu:
Will merge for now then add `triggerWith`

thgrund:
That sounds good to me. I also think that a `triggerWith` is certainly a good solution. Especially since various `with` functions already exist. And please excuse me for not responding more here. I did not have this thread in mind anymore.

PR #738, opened 2020-12-14
Handle ? after ! and @_

### LINKED ISSUE #646: Syntax error in mini-notation with ! and ?
**Describe the bug**
`d1 $ s "bd!8?"` produces:
```
Error in pattern: Syntax error in sequence:
  "bd!8?"
       ^  
unexpected '?'
expecting charnum, rest, "[", "{", "<", "." or end of input
```
Would probably have expected it to behave like this, which works:

`d1 $ s "[bd!8]?"`

ndr-brt:
Fix #646


ndr-brt:
Fixed the issue also for `@`.
Also removed some `spaces` that were marked with a TODO. 
Tests were still green so I guess that was ok.

jwaldmann:
I can confirm that `spaces` is not needed after `angles` and the like, since the standard lexers remove whitespace after tokens (that's why LanguageDef contains the definition of whitespace). Stand-alone test case:
```
ghci -package parsec
import Text.Parsec
import Text.Parsec.Token
import Text.Parsec.Language
lexer = makeTokenParser haskellDef

parseTest ((many $ angles lexer $ natural lexer) <* eof) "< 123 >   <456> "

   ==>  [123,456]
```

yaxu:
Thanks!

Issue #740, opened 2020-12-17
Support SuperDirt control buses


yaxu:
See discussion here -> https://github.com/musikinformatik/SuperDirt/issues/190

yaxu:
duplicate of #762

PR #741, opened 2020-12-17
Bus control - superdirt handshaking and on-the-fly event effects

yaxu:
Ref #740 and https://github.com/musikinformatik/SuperDirt/issues/190


telephon:
Let me know if there is anything to do on the superdirt side …

telephon:
Superdirt could also send information about all synth parameters, btw. if that is useful …

telephon (approved):

Issue #745, opened 2020-12-30
(with tidal-1.6.2) n vs. note


jwaldmann:
(I will fill in more info, for now it's just a reminder)

https://club.tidalcycles.org/t/with-tidal-1-6-2-n-vs-note/2683

perhaps it's two bugs?
```
s "superpiano" >| n "[c d e f]"

(0>¼)|n: 0.0f, s: "superpiano"
(¼>½)|n: 2.0f, s: "superpiano"
(½>¾)|n: 0.125f, s: "superpiano"
(¾>1)|n: 0.2f, s: "superpiano"
```
and
```
s "superpiano" >| note "[c d e f]"
...
Failed to send. Is the 'SuperDirt' target running? Maybe.fromJust: Nothing
CallStack (from HasCallStack):
  error, called at libraries/base/Data/Maybe.hs:148:21 in base:Data.Maybe
  fromJust, called at src/Sound/Tidal/Stream.hs:217:39 in tidal-1.6.2-7f2a9abbae25afcc64d0a7d96ffe51eb55482b087b15e92\
7fbaba2cc55015df2:Sound.Tidal.Stream
```

ndr-brt:
I tried the second pattern with a test:
```
        it "should parse correctly" $ do
          compareP (Arc 0 1)
            (sound "superpiano" >| note "[c d e f]")
            (sound "superpiano" >| note "[0 2 4 5]")
```

and it's working. have you build tidal with the last sources?

the problem with `n` exists

jwaldmann:
yes the test is working (set of events is equal) but the sending to the back-end is not.

ndr-brt:
interesting.
`n` is not working as expected because:
https://github.com/tidalcycles/Tidal/blob/4c79a3314dacd77d3e1327a3cc54361da8dffeec/src/Sound/Tidal/Params.hs#L284
should be a Pattern Note

ndr-brt:
> yes the test is working (set of events is equal) but the sending to the back-end is not (try `d1 $ ...`) -- but wait, I will re-check.

maybe because the VN pattern is missed from `toDatum`?
https://github.com/tidalcycles/Tidal/blob/4c79a3314dacd77d3e1327a3cc54361da8dffeec/src/Sound/Tidal/Stream.hs#L209

But I still don't understand why tests are working

jwaldmann:
test are working because they compare this:
```
sound "superpiano" >| note "[c d e f]"
[...]

PR #746, opened 2020-12-30
Add VN to toDatum patterns

ndr-brt:
Add VN pattern to datum, as discussed in #745 
Add a simple test, not so meaningful but needed to replicate the bug and write the missing code


yaxu:
Thanks!

PR #749, opened 2021-01-06
fix #748

### LINKED ISSUE #748: tidal-see-output called by tidal-start-haskell displays duplicate buffer
tidal_version is 1.6.1

If I type C-c C-s (bound to tidal-start-haskell) starting from a single buffer displaying a Tidal file (first screenshot) I end up with three buffers (two duplicates of the original buffer plus the Tidal interpreter buffer (second screenshot)). I believe the desired outcome would be to have only one Tidal file buffer alongside the Tidal interpreter buffer (third screenshot).

I am on MacOS Big Sur 11.1  and Emacs 27.1.

Screen Shot 2021-01-03 at 14 17 07

Screen Shot 2021-01-03 at 14 17 24

Screen Shot 2021-01-03 at 14 17 35

jwaldmann:
as discussed in the issue. removing this line seems to solve the problem.

### DIFF
--- a/tidal.el
+++ b/tidal.el
@@ -118,7 +118,6 @@
   (interactive)
   (when (comint-check-proc tidal-buffer)
     (delete-other-windows)
-    (split-window-vertically)
     (with-current-buffer tidal-buffer
       (let ((window (display-buffer (current-buffer))))
 	(goto-char (point-max))


yaxu:
Thanks again

Issue #752, opened 2021-01-07
Making the first argument of binaryN patternable


fracnesco:
I think it would be nice to make the number of bits in binaryN patternable to allow something like this

binaryN "<8 [8 16]>" 2135890

yaxu:
Thanks @ndr_brt

PR #754, opened 2021-01-07
Binaryn patternable

ndr-brt:
Implements #752

### DIFF
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -1892,7 +1892,7 @@ _binaryN :: Int -> Pattern Int -> Pattern Bool
 _binaryN n p = squeezeJoin $ _binary n <$> p
 
 binaryN :: Pattern Int -> Pattern Int -> Pattern Bool
-binaryN n p = squeezeJoin $ (\np -> _binaryN np p) <$> n
+binaryN n p = tParam _binaryN n p
 
 binary :: Pattern Int -> Pattern Bool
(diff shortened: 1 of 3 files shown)


src/Sound/Tidal/UI.hs
```diff
@@ -1948,8 +1948,11 @@ __binary n num = map (testBit num) $ reverse [0 .. n-1]
 _binary :: Data.Bits.Bits b => Int -> b -> Pattern Bool
 _binary n num = listToPat $ __binary n num
 
-binaryN :: Int -> Pattern Int -> Pattern Bool
-binaryN n p = squeezeJoin $ _binary n <$> p
+_binaryN :: Int -> Pattern Int -> Pattern Bool
+_binaryN n p = squeezeJoin $ _binary n <$> p
+
+binaryN :: Pattern Int -> Pattern Int -> Pattern Bool
+binaryN n p = squeezeJoin $ (\np -> _binaryN np p) <$> n
```
yaxu, line 1895: I'd normally use `innerJoin` rather than `squeezeJoin`, using the tParam shortcut:
```
 binaryN = tParam _binaryN
```
I think `squeezeJoin` makes sense for the `_binaryN` definition though.

ndr-brt: Yes, thanks for the feedback, I'm trying to learn the whole `join` stuff. 
Did the fix

yaxu:
Thanks! lmk if you've questions about the joins

PR #759, opened 2021-01-14
Makes `whenmod` work with a non-integer numbers of cycles

th-four:
This allows things like `whenmod 8.5 6.5`, which was previously not possible due to `mod` only working on integers.

### DIFF
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -551,8 +551,8 @@ as dense:
 d1 $ whenmod 8 4 (density 2) (sound "bd sn kurt")
 @
 -}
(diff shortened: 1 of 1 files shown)


yaxu:
Thanks! Probably the build failure is caused by tidal-parse needing changes to match, I can look at that

yaxu:
This seems to be doing more like I expected:
```
whenmod :: Time -> Time -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
whenmod a b = Sound.Tidal.Core.whenT (\t -> ((t `mod'` a) >= b ))
```
Working with time as rational rather than integral

PR #764, opened 2021-01-22
New Sound.Tidal.Params.Standard

yaxu:
With generated parameters


src/Sound/Tidal/Params/Generate.hs
```diff
@@ -0,0 +1,391 @@
...
+        ]
+
+genericParams :: [(String, String, String)]
+genericParams = [
+  ("s", "toArg", "for internal sound routing"),
+  ("f", "from", "for internal sound routing"),
+  ("f", "to", "for internal sound routing"),
+  ("f", "accelerate", "a pattern of numbers that speed up (or slow down) samples while they play."),
+  ("f", "amp", "like @gain@, but linear."),
+  ("f", "attack", "a pattern of numbers to specify the attack time (in seconds) of an envelope applied to each sample. Only takes effect if `release` is also specified."),
```
telephon, line 174: > Only takes effect if `release` is also specified.
Not true, it does (should) take effect alone.

src/Sound/Tidal/Params/Generate.hs
```diff
@@ -0,0 +1,391 @@
...
+  ("f", "panspan", "a pattern of numbers between -inf and inf, which controls how much multichannel output is fanned out (negative is backwards ordering)"),
+  ("f", "pansplay", "a pattern of numbers between 0.0 and 1.0, which controls the multichannel spread range (multichannel only)"),
+  ("f", "panwidth", "a pattern of numbers between 0.0 and inf, which controls how much each channel is distributed over neighbours (multichannel only)"),
+  ("f", "panorient", "a pattern of numbers between -1.0 and 1.0, which controls the relative position of the centre pan in a pair of adjacent speakers (multichannel only)"),
+  ("f", "pitch1", ""),
+  ("f", "pitch2", ""),
+  ("f", "pitch3", ""),
+  ("f", "portamento", ""),
+  ("f", "rate", "used in SuperDirt softsynths as a control rate or 'speed'"),
+  ("f", "release", "a pattern of numbers to specify the release time (in seconds) of an envelope applied to each sample. Only takes effect if `attack` is also specified."),
```
telephon, line 248: >Only takes effect if `attack` is also specified.

No, should work without. `hold` is the only exception, which is reasonable.

yaxu:
Not sure why stack isn't building this..

Issue #743, opened 2020-12-26
get location of BootTidal.hs


jwaldmann:
Hi. 

This is another way of finding `BootTidal.hs`, avoiding gymnastics with `ghc-pkg field` and `ghc --numeric-version`. Might help with #583 #615. Worksforme, and I can make a PR.

Put this in `tidal.cabal`
```
...
  autogen-modules: Paths_tidal
  
  Exposed-modules:     Paths_tidal
                       Sound.Tidal.Bjorklund
...
```
then after installing (`cabal install --lib`)
```
ghc -e "import Paths_tidal" -e "getDataDir"

==> "/home/waldmann/.cabal/store/ghc-8.10.2/tidal-1.6.2-7f2a9abbae25afcc64d0a7d96ffe51eb55482b087b15e927fbaba2cc55015df2/share"
```
So  put these lines in `tidal.el`:
```
          ((or (string-equal system-type "darwin") (string-equal system-type "gnu/linux"))
           '(("path" . "echo -n data-dir: && ghc -e 'import Paths_tidal' -e 'getDataDir>>=putStr'")
             ("separator" . "/")
             ))
```
where I need to add "data-dir:" artifically but the next part of the code expects it.

I like this because it is just calling `ghc`, and `ghci` is needed anyway for running tidal.
(Current solution calls `ghc-pkg` and runs into its inconsistent (with `ghc(i)`) handling of env files.)

Features used:
* Cabal access to data files https://cabal.readthedocs.io/en/3.4/cabal-package.html#accessing-data-files-from-package-code
* ghc expression evaluation https://downloads.haskell.org/ghc/latest/docs/html/users_guide/using.html#eval-mode (but using `-e` two times seems undocumented, cf. https://gitlab.haskell.org/ghc/ghc/-/issues/19122)

One could probably avoid the extra call of `ghc` and instead load the boot file as the very first command of Tidal's ghci session. This would require some contraption - does ghci allow bash-like "HERE scripts"? [EDIT] not HERE scripts but "command substitution" https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html#Command-Substitution

yaxu:
Looks good! I only know about :script for loading scripts and :{ and :} for denoting blocks of code in ghci.

jwaldmann:
I can make a PR.

It should be tested for other ways of installing tidal (and ghc) -- I am only doing cabal (user) install from source.

[...]

PR #771, opened 2021-01-30
ignore noise on stderr produced by ghc-8.6 and ghc-8.8

jwaldmann:
https://github.com/tidalcycles/Tidal/issues/743#issuecomment-769681226


yaxu:
Thanks!

Issue #768, opened 2021-01-26
Move the CI to github actions


ndr-brt:
A message appeared on travis:
> Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.

Made me think about the CI process on the atom-tidalcycles plugin.
We switched pretty seamlessly (https://github.com/tidalcycles/atom-tidalcycles/issues/142)

What if we swtich it also on Tidal?

ndr-brt:
this could be a good starting point:
https://github.com/haskell/actions

ndr-brt:
Opened a PR, tried that on my fork: https://github.com/ndr-brt/Tidal/actions/runs/530729880
The time to execute 9 jobs is 3m35s.
The first time it will take a little more because the caches are empty.

There's some questions I have for everyone:
* there's the possibility to run the build/test process on different environments, do we need that or ubuntu is enough?
* do we need the two "Allowed Failures" builds? I did not include that because I don't see the point of those
* do we still need to test against older ghc version like 8.0.2? What's the minimum ghc acceptable version?

yaxu:
I think ubuntu is probably enough, we don't want to waste electricity on other linux versions at least. 
I'm not sure where the 'allowed failures' come from and agree we should drop them! They have always failed as far as I remember.
The oldest ghc installs are probably from debian stable. The current stable has been on ghc 8.4.4 since July 2019. Before that it was ghc 8.0.1 on stretch. I think there will be people with laptops still running stretch, and probably this is the version that is most likely to have install problems, so I think it's worth keeping the 8.0.x build.

In summary I think that's great as-is!

ndr-brt:
Ok, added the 8.0.2 build and cleaned up the workflow file. I think it's ready to be merged

yaxu:
Nice one!

PR #774, opened 2021-02-02
Define workflow on github actions

ndr-brt:
As discussed in #768


yaxu:
Looks good! Let me know if I need to do something..

ndr-brt:
@yaxu I put some questions on the issue: #768

yaxu:
Thanks very much!

PR #776, opened 2021-02-05
Mini notation trailing and leading whitespaces

### LINKED ISSUE #772: Mini-notation spaces less forgiving
A couple of user reports:

This used to work (spaces around the `%`):
```
"{a b c} % 4" :: Pattern String
```
As did this:
```
" bd" :: Pattern String
```
I think these are probably regressions, following from here:
https://github.com/tidalcycles/Tidal/commit/170221292819e4157e9ac46334c17a4df1e52bc4

I remember some discussion with @ndr-brt and @jwaldmann about this but can't find it at the moment..

ndr-brt:
Fix #772


yaxu:
Super

PR #779, opened 2021-02-12
better error reporting and some new definitions available in tidal-parse

dktr0:


dktr0:
Failing tests are likely just because I just uploaded haskellish-0.2.4 to hackage now (forgot while working earlier because Estuary builds with nix, bypassing hackage)

yaxu:
OK have set the tests to rerun

dktr0:
The test for stack lts-16... is failing because, for no reason I understand, it is grabbing an *older* version of Haskellish and saying it is because allow-newer is enabled... "WARNING: Ignoring tidal-parse's bounds on haskellish (>=0.2.4.2 && <0.3); using haskellish-0.2.3.1.
Reason: allow-newer enabled."
Is there anything in the way the CI is setup that would be causing this? I can't think of anything else, anyway.

yaxu:
That's strange. Haskellish isn't even in stackage, afaict. So I don't see why it would install an old version

dktr0:
Something to do with stack falling back on out of date state-of-Cabal, when it realizes the package is not in stackage?

yaxu:
Do the other stack builds work? I'm not sure if it makes sense to treat anything other than building against nightly as a failure and maybe most recent lts anyway.

yaxu:
I guess this is like trying to install a new .deb in a distro that's several years old.. Stackage is a curated set of package versions that work together, if someone wants a new tidal they install it from the latest lts or nightly release.

dktr0:
I haven't looked at the stack configurations for Tidal (yet), but an option could be to not use allow-newer. I'm wondering if with allow-newer when it can't find the package it uses the one that was the latest at the time the lts snapshot was taken. Since the whole point of using stack is to pin very specific versions of packages, allow-newer sort of defeats that purpose, even if it is occasionally necessary... The other stack builds are just "cancelled" - is that some kind of knock-on effect from the lts-16... one failing?

yaxu:
I'll merge for now and tidy up on main

yaxu:
The error message still doesn't make sense, but this PR on haskellish might fix things @dktr0 : https://github.com/dktr0/Haskellish/pull/1

PR #783, opened 2021-02-18
Fix parsing of pattern groupping shorthand

wattroll:
Regression was introduced in 262e082cd5c33c824452d93fdcf972d4ce12ca56 during code reformating as a part of PR #776.

This caused basic patterns like `a . b` fail rather confusingly:

    Error in pattern: Syntax error in sequence:
      "a . b"
         ^
    unexpected " "
    expecting ".."

This commit brings back the forgotten `try` for `symbol ".."` and adds parsing tests for a pattern group and its shorthand form.

---

We should extend the parser test suite with the rest of the examples from https://tidalcycles.org/Mini_notation_syntax to avoid such regressions. Being new to Tidal's microsyntax it took me a lot of squinting to establish that the parser is indeed broken and it's not something with my editor/terminal setup that causes basic examples to fail.

### DIFF
--- a/src/Sound/Tidal/ParseBP.hs
+++ b/src/Sound/Tidal/ParseBP.hs
@@ -339,7 +339,7 @@ pSequence f = do
     a <- pPart f
     spaces
     do
-      symbol ".."
+      try $ symbol ".."
       b <- pPart f
       return $ TPat_EnumFromTo a b
       <|> pElongate a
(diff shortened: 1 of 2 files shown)


yaxu:
Thanks very much, and agreed we should focus on improving the test suite before major changes to the parser. Taking all the examples in the documentation and testing against their current outputs would be a good start I think.
I think we should release a new tidal with this fix once the stack build is fixed.

PR #788, opened 2021-02-19
tidal-parse: updated dependency bounds; ControlMap replaced with Valu…

dktr0:
…eMap

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -7,6 +7,6 @@ packages:
 extra-deps:
   - hosc-0.18.1
   - microspec-0.2.1.3
-  - haskellish-0.2.3.1
+  - haskellish-0.2.4.3
 
 allow-newer: true
--- a/tidal-parse/src/Sound/Tidal/Parse.hs
+++ b/tidal-parse/src/Sound/Tidal/Parse.hs
@@ -9,7 +9,7 @@ import           Data.Bifunctor
 import           Control.Monad.Except
 import qualified Data.Text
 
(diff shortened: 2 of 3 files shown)


dktr0:
This incorporates the wider bounds on template-haskell - same failing test result on the stack build as before with the stack build grabbing a much older version of Haskellish. I'm wondering if this would succeed without "allow-newer" in the stack configuration. Is there a good reason allow-newer is enabled?

dktr0:
oh wait now I think I see it - the specific (old) version of haskellish is named in the stack configuration!

dktr0:
Have updated - fingers crossed...

dktr0:
I remember now: anything that isn't in a stack snapshot MUST be referenced by a specific pinned version.

dktr0:
Hurray - now the stack one builds.

yaxu:
Great thanks! Maybe it's worth putting haskellish in stackage? If I remember right, it's just a case of listing it with a maintainer in this file: https://github.com/commercialhaskell/stackage/blob/master/build-constraints.yaml

I'm actually listed as the maintainer of hosc, because Rohan didn't want the bother, so could take haskellish there too if you prefer.

On the other hand might not gain a huge amount from having haskellish in stackage at the moment.

By the way sorry for breaking tidal-parse, will try to remember to test changes there in the future.

dktr0:
I'll probably do this sometime soon. I'm not sure it will help economize tidal-parse work that work - I kind of expect there will often be reasons to pin specific versions, as Haskellish evolves (not exactly rapidly, but not exactly stable/old hat either...).

Issue #797, opened 2021-02-25
mute/unmute/unmuteall with midi


ndr-brt:
Would be nice to have the mute/unmute/unmuteAll functions controllable by midi (through osc)
As proposed in https://club.tidalcycles.org/t/the-atom-plugin-thread/2244/57

yaxu:
Ref #682

mindofmatthew:
I've added a PR that would add this functionality via `/mute`, `/unmute`, `/muteAll`, and `/unmuteAll` messages sent to a stream's control port, mirroring the relevant functions.

I'm not particularly wedded to those command paths—not sure if there's a current place for OSC api discussion, but thought I'd go ahead and implement some functionality as a starting point.

yaxu:
I think these osc paths are fine good. I think we just need the sc example expanding to include this now

PR #800, opened 2021-03-02
Add OSC API for muting/unmuting a stream

mindofmatthew:
Ref #797

### DIFF
--- a/src/Sound/Tidal/Stream.hs
+++ b/src/Sound/Tidal/Stream.hs
@@ -647,6 +647,7 @@ ctrlResponder waits c (stream@(Stream {sListen = Just sock}))
             bufferIndices [] = []
             bufferIndices (x:xs') | x == (O.ASCII_String $ O.ascii "&controlBusIndices") = catMaybes $ takeWhile isJust $ map O.datum_integral xs'
                                   | otherwise = bufferIndices xs'
+        -- External controller commands
         act (O.Message x (O.Int32 k:v:[]))
           = act (O.Message x [O.string $ show k,v])
         act (O.Message _ (O.ASCII_String k:v@(O.Float _):[]))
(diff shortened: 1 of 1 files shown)


src/Sound/Tidal/Stream.hs
```diff
@@ -655,6 +656,19 @@ ctrlResponder waits c (stream@(Stream {sListen = Just sock}))
           = add (O.ascii_to_string k) (VS (O.ascii_to_string v))
         act (O.Message _ (O.ASCII_String k:O.Int32 v:[]))
           = add (O.ascii_to_string k) (VI (fromIntegral v))
+        -- Stream playback commands
+        act (O.Message "/mute" (O.Int32 k:[]))
+          = streamMute stream (fromIntegral k)
```
yaxu, line 661: This throws a compile warning because the second parameters to `streamMute` takes anything that can be passed to `show`, and `fromIntegral` can produce more than one type, so this is ambiguous. Actually do you need a `fromIntegral` there? I'd have thought `k` would be in the `Show` class.

src/Sound/Tidal/Stream.hs
```diff
@@ -655,6 +656,19 @@ ctrlResponder waits c (stream@(Stream {sListen = Just sock}))
...
+        act (O.Message "/mute" (O.ASCII_String k:[]))
+          = streamMute stream (O.ascii_to_string k)
+        act (O.Message "/unmute" (O.Int32 k:[]))
+          = streamUnmute stream (fromIntegral k)
```
yaxu, line 665: Same here

src/Sound/Tidal/Stream.hs
mindofmatthew: Oh, yeah, that's totally right—not sure how I didn't notice the warning

yaxu:
Thanks!

PR #804, opened 2021-03-04
Fix indentation on some UI tests

mindofmatthew:
I was looking to add a UI test and noticed that several test cases are mistakenly interpreted as sub-cases of arpeggiate


yaxu:
Thanks!

PR #805, opened 2021-03-06
Fix issue with OSC path substitution

mindofmatthew:
I was adding some tests and discovered a bug with custom OSC paths: If you specify a path with a fallback (like `"/play/{s=bd}"`), Tidal always subs in the fallback, even if the param name exists in a given event.


yaxu:
Aha, thanks!

Issue #803, opened 2021-03-04
Pattern ID Types


mindofmatthew:
Because patterns can be referred to by any instance of `Show`, it complicates the process of defining an OSC api for accessing them. Would it make sense to change pattern IDs to work more similarly to external control IDs, where it only accepts a String or an Integer and `p 1` replaces the same pattern as `p "1"`?

yaxu:
Yes that makes sense to me

mindofmatthew:
Sounds good! Would it make sense to extend the logic in the other end, so that `cF0 20` works within Tidal, or is that not worth doing right now?

yaxu:
Hm, I can't think of a good reason not to.

PR #807, opened 2021-03-17
Pattern IDs

mindofmatthew:
Here's my proposed modification to pattern IDs in line with #803. This would specify that patterns can only be identified by an integer or a string (this could be expanded to include other types that would be useful IDs, but ints and strings were the ones I was aware of being used in practice).

This approach has two benefits:
- This ensures that all IDs that can be created in code can be easily referenced via OSC.
- This now allows for overloaded implementations of `mute` etc, that can take either a single ID or a list of IDs for muting multiple channels. `mute 1` `mute "bass"` `mute [2..6]` and `mute ["drum", "bass"]` are now all legal, and the OSC mirror this behavior by accepting multiple arguments.

I was hoping there was some sort of clever abstract typeclassing I could do to prevent the ID module from being so verbose, but I couldn't come up with an approach that didn't create some ambiguity that GHC couldn't resolve. Happy to hear if you've got thoughts, or if you'd prefer that I do anything different with the code in terms of naming, formatting, etc.


mindofmatthew:
Ah, I guess this still isn't working, I think because I'm still relying on some type inference that GHCi adds. I'll keep working on it.

yaxu:
(sorry marked as 'ready to review' with a stray click..)

mindofmatthew:
Okay, @yaxu, the simplified version of this change is ready for review! Let me know if there's anything here that you want changed.

On compatibility, this is technically a breaking change for anyone who's used `Show`ables other than ints and strings, but given that those are the only two types ever suggested in documentation, it might be okay to ignore the incompatibility there?

Fixes #807. Also, maybe #846, but I'm not sure if there's anything specific to tidal-listener that's still captured by that issue.

PR #809, opened 2021-03-18
Ctrl UI Changes

mindofmatthew:
(Whittling away at my other PR while I figure out how to proceed with it.)

It seems to me that it would be good to specifically pattern match on the `"/ctrl"` OSC path, especially if Tidal moves toward a fuller OSC api. This is technically a breaking change, but seems like it would be low-impact, because all the documentation I've seen has consistently referred to that message as `/ctrl`. Let me know if that's not the case.

I also added a handy function for getting a current state variable, which was nice for testing. I also made a BootTidal alias. `get` seemed to match `setI` etc, but maybe it's too generic? Let me know what you think—thanks!

### DIFF
--- a/BootTidal.hs
+++ b/BootTidal.hs
@@ -66,7 +66,7 @@ let only = (hush >>)
 :}
 
 :{
-let get = streamGet tidal
+let getState = streamGet tidal
     setI = streamSetI tidal
     setF = streamSetF tidal
     setS = streamSetS tidal
--- a/BootTidal.hs
+++ b/BootTidal.hs
@@ -66,7 +66,8 @@ let only = (hush >>)
 :}
 
 :{
-let setI = streamSetI tidal
+let get = streamGet tidal
+    setI = streamSetI tidal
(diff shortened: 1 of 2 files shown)


yaxu:
Yes I'm surprised that `/ctrl` isn't already pattern matched. I'm also a bit wary about using `get` for something. Could such a short word be used for something that could get more use? I can't think of anything immediately.. But maybe something a bit more specific like `getState` would be better?

(There is already `getI` and friends for `Value -> Maybe Int`, but I'd sa those are really internal functions and could be renamed or hidden.)

mindofmatthew:
Yes, `getState` sounds good. I also considered `getCtrl`, but it didn't seem in line with the direction that the language around that state is going. I can go ahead and push that change.

(Incidentally, I wonder if `toI` or maybe `iFromValue` would be good symmetrical names with `toValue`, since those seem to be inverse operations. But that's outside the scope of this PR.)

yaxu:
Thanks!

PR #815, opened 2021-04-02
Add more playback controls to OSC API

mindofmatthew:
This PR adds a missing unsoloAll to the set of stream playback controls. It also adds support for the rest of the similar playback controls (solo/unsolo/unsoloAll/hush) in the OSC control API.


yaxu:
Thanks!

PR #816, opened 2021-04-02
Disable busses for MIDI parameters

mindofmatthew:
Following up on @thgrund's thread on the forum about MIDI CC busses, this adds all of the SuperDirt MIDI params to the `nobus` list because busses are only for SC synth params at the moment.


thgrund:
This looks good thank you very much! And also nice to see that the contributions are growing :-) (or does it just seem to be for me? ^^)

yaxu:
Thanks!

PR #818, opened 2021-04-06
Test and bugfix basic patterns (sine, etc)

mindofmatthew:
I noticed a couple of oversights with how the bipolar continuous patterns are defined. I also added some tests of those functions, partially to evaluate the fixes and partially to figure out how to write Quick Check properties.

Let me know if there's anything that needs tweaking!


yaxu:
This looks great, thanks for catching this!

PR #819, opened 2021-04-07
.cabal: tweak description

juhp:
tiny changes to the description field in tidal.cabal:

- make "live coding pattern" plural
- move it up to usual place in .cabal file

### DIFF
--- a/tidal.cabal
+++ b/tidal.cabal
@@ -1,7 +1,7 @@
 name:                tidal
 version:             1.7.3
 synopsis:            Pattern language for improvised music
--- description:
+description:         Tidal is a domain specific language for live coding patterns.
 homepage:            http://tidalcycles.org/
 license:             GPL-3
 license-file:        LICENSE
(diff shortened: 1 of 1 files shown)


yaxu:
Hm, what's the motivation for this? I think I prefer 'pattern'.

juhp:
Okay, the plural thing was just a suggestion.
You know best of course: just "DSL for live coding pattern" doesn't read so well for me,
but maybe it is familiar to people already. I am new to live coding.

The second part was just moving the description up in the .cabal file.

Reducing the compiler warnings might be a more useful contribution. :-)

yaxu:
You're not alone, I have had similar edit suggestions elsewhere. It'd be interesting if a passing linguist could clarify the difference between "making pattern" and "making patterns", I'm scratching my head over it..

juhp:
Well in "for making patterns" we have a continuous verb (gerund) acting on a noun.
I don't know how to parse "for making pattern" maybe it is treating "making pattern" as a type of verb?? I am not sure

Then should it be "to live coding pattern"??

yaxu:
(I've merged, but..)
I guess I am thinking of 'pattern' as an activity by itself, like 'live coding dance', which seems better than 'live coding dances' to me?

juhp:
I am totally fine if you want to change it back later - just a small point I spotted while reviewing the package for Fedora and then trying to cleanup the .cabal file a tiny bit. :-)

I think "Tidal is a domain specific language to live coding pattern" could be better in this case.

PR #822, opened 2021-05-07
Handle negative ratio shorthands also for rationals

ndr-brt:
I noticed that something like
```
d1 $ off "-e" $ s "bd"
```
was still not working.


yaxu:
Thanks!

PR #823, opened 2021-05-10
Chord test cases

cleary:
first pass at some chord test cases - 

covers modifiers at this point (length mods, inversions and open voicing)

todo: 
 - test some random standard chords with no modifiers?


yaxu:
Thanks!

PR #829, opened 2021-05-25
Ratio multiply

### LINKED ISSUE #827: ratio shorthand multiplication does not behave as expected
this code 
```
s "bd" # speed "3h"
```

when evaluated prints:
```
(0>½)-1|s: "bd", speed: 3.0f
0-(½>1)|s: "bd", speed: 0.5f
```

i was expecting something like this:
```
(0>1)|s: "bd", speed: 1.5f
```

ndr-brt:
Fix #827 . 
Hope that's the expected behaviour :)


yaxu:
Thanks!

PR #831, opened 2021-05-28
fixes #830

### LINKED ISSUE #830: splice overrides speed
I'd expect this to reverse the slices:
```
splice 8 "0 1 2 3 4 5 6 7" # s "breaks125" |* speed "-1"
```
However the speed has no effect. It seems `splice` is overwriting any existing `speed` value, it would probably be better if it multiplied an existing value where present?

onthepeakofnormal:
minor change to _splice so that it respects if the speed parameter was 
already declared

### DIFF
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -285,7 +285,9 @@ randslice = tParam $ \n p -> innerJoin $ (\i -> _slice n i p) <$> _irand n
 
 _splice :: Int -> Pattern Int -> ControlPattern -> Pattern (Map.Map String Value)
 _splice bits ipat pat = withEvent f (slice (pure bits) ipat pat) # P.unit (pure "c")
-  where f ev = ev {value = Map.insert "speed" (VF d) (value ev)}
+  where f ev = case Map.lookup "speed" (value ev) of
+                        Nothing -> ev {value = Map.insert "speed" (VF d) (value ev)}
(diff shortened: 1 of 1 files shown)


src/Sound/Tidal/Control.hs
```diff
@@ -285,7 +285,9 @@ randslice = tParam $ \n p -> innerJoin $ (\i -> _slice n i p) <$> _irand n
 
 _splice :: Int -> Pattern Int -> ControlPattern -> Pattern (Map.Map String Value)
 _splice bits ipat pat = withEvent f (slice (pure bits) ipat pat) # P.unit (pure "c")
-  where f ev = ev {value = Map.insert "speed" (VF d) (value ev)}
+  where f ev = case Map.lookup "speed" (value ev) of
+                        Nothing -> ev {value = Map.insert "speed" (VF d) (value ev)}
```
yaxu, line 289: This is causing a warning, it'll fail on a non floating point `speed` value. That's unlikely to happen, but worth swapping the two lines around, and replacing `Nothing` with `_` as the default case.

yaxu (changes requested):

src/Sound/Tidal/Control.hs
onthepeakofnormal: Ah yes that makes sense! Thanks!

yaxu:
Thanks!

PR #833, opened 2021-06-10
Fail parsing floating point to int

### LINKED ISSUE #832: Dot shorthand in integer patterns
This is not strictly a bug, but a behaviour that I found very confusing.

Today I was doing something like this:
```
fast 2
$ note (toScale' 53 [0,9,14,22] 
$ (|* "<1 2>") 
"0 1 2 3 4")
```
This worked ok. But then for some reason, I tried to multiply by a fractional. It shouldn't have worked, since the function in line 3 was being applied to a pattern of integers, but it did:

```
fast 2
$ note (toScale' 53 [0,9,14,22] 
$ (|* "<1 1.5>") 
"0 1 2 3 4")
```

However, the results were very strange, and it took me some time to understand what was going on. If I undestand correctly, when summing/multiplying/etc a pattern of integers by a number with a dot in it (even an integer, like `1.0`), this is interpreted not as a number but as the dot shorthand for the mininotation. This means that `1.5` becomes `[1][5]`.

I think this is very unintuitive behaviour. `1.0` has the same meaning everywhere, except where it doesn't and it becomes `[1][0]`, which it is probably not expected to. The most reasonable behaviour was to just interpret the number as a real/fractional and throw a type error, instead of giving it another chance as something else entirely.

Imho, the best way to solve this would be to always interpret numbers with dots in it as real/fractional, and to require a space between numbers when one wants to use the dot shorthand with numbers. So the `1.5` in my code would throw a type error, while `1 . 5` would be valid.

ndr-brt:
Fix #832

### DIFF
--- a/src/Sound/Tidal/ParseBP.hs
+++ b/src/Sound/Tidal/ParseBP.hs
@@ -471,12 +471,6 @@ parseIntNote = do s <- sign
                 where
                   isInt x = x == fromInteger (round x)
 
--- TODO: this seems to be unused
-parseInt :: MyParser Int
-parseInt = do s <- sign
-              i <- integer
-              return $ applySign s $ fromIntegral i
-
 pIntegral :: Integral a => MyParser (TPat a)
 pIntegral = wrapPos $ do i <- parseIntNote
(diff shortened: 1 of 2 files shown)


yaxu:
Thanks!

PR #836, opened 2021-06-20
fixing broken example for contrast

mxmxyz:
The example in the code for the `contrast` function was broken because the arguments weren't in the correct order. This seems to fix it.


yaxu:
Thanks!

PR #840, opened 2021-09-22
Consistent naming cont'd

cleary:
Hi @yaxu, 
I've added a few more fixes into the naming that I missed originally -

I've separated a commit for consideration, which is to allow the use of `'M` (capital M) as a Major shorthand. This hasn't been applied to any other Major chords though - keen for your thoughts, and happy to discard.


cleary:
@yaxu I'm unsure what the check failures mean, is there a specific problem I need to deal with to get this accepted?

yaxu:
Hi @cleary, sorry I've been distracted the last month. I will have a bit more time from November and all the time in the world from December!

I think the test failures are probably unrelated to your changes, I've re-run them just now.

Looking at your changes, I think they break backward compatibility? Perhaps we could copy the old definitions to a separate part of the file and mark them as deprecated for a little while, or something like that?

Issue #850, opened 2021-10-06
tidal-listener: add workflows to build up binaries


ndr-brt:
Like @polymorphicengine did on their fork: https://github.com/polymorphicengine/tidal-listener/tree/master/.github/workflows

Issue #315, opened 2018-07-11
Build a Tidal binary 


yaxu:
It'd be great to make a Tidal binary so that people don't have to install Haskell and all the libraries. Currently if some hackage package or new haskell version has an issue (and this happens fairly often), then Tidal becomes very difficult to install.
I see this happening in two steps:

* Step 1 - make a drop-in replacement for ghci based on hint, that has tidal etc loaded. Good enough for the current editor plugins to use
* Step 2 - make something like a websocket api that can make all kinds of editor interactions possible

Step 2 is [already kind of done](https://github.com/yaxu/tidal-websocket), but using it would require updating the editor plugins, and the websocket library has some dependencies that might complicate making a standalone binary of tidal-websocket. So I think stripping things back and getting step 1 done is a good first aim.
We have three problems - linux, mac os x and windows. I'll be focusing on Linux, if anyone fancies helping with the other platforms that'd be great.
There are [simple instructions for packaging binaries](http://gelisam.blogspot.com/2014/12/how-to-package-up-binaries-for.html). The complication is that it's fiddly to get '[hint](http://hackage.haskell.org/package/hint)', the haskell interpreter library, to work without the whole of ghc being installed. However, [it is possible](https://github.com/gelisam/deploy-hint/tree/85918a527c8801f5b56f94897e6744c2cc86ee38) ([more info](https://github.com/haskell-hint/hint/issues/3)).

yaxu:
Work in progress over here:
  https://github.com/tidalcycles/tidali

yaxu:
Giving up on this.

yaxu:
(making a tidal-parse library a better approach, I think)

yaxu:
This looks v promising:
https://mpickering.github.io/ide/posts/2020-07-10-ghc-libdir.html

tonyday567:
Having dug fairly deeply into the haskell ecosystem, I think that ghcup is an interesting project that could be leveraged to provide a hassle-free tidal install. Adopting a similar style, you would see something like this:

```
✗  tidal 1.6.1 compatible with SuperCollider 2.3.7
✓  tidal 1.7.4 latest, recommended, compatible with 3.11.1
✓  ghc 8.10.4 recommended
✓  stack 1.9.1 recommended
✓  SuperCollider 3.11.1
...

that wraps the whole install experience.  Call it tidalup, say, so as not to confuse people, but just really piggy back the projects technology to be able to install everywhere.

[...]

PR #852, opened 2021-10-17
Add listener build workflows

ndr-brt:
Ref #850 and #315 
Adapted from https://github.com/polymorphicengine/tidal-gui
They will run only after a tag with the format `*.*.*`, so only after a tidal release.
Tested on my fork.

### DIFF
--- a/.github/workflows/listener-build-linux.yml
+++ b/.github/workflows/listener-build-linux.yml
@@ -1,0 +1,126 @@
+name: build-listener-linux
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ghc: ['9.0.1']
+        cabal: ['3.4.0.0']
+        
+    steps:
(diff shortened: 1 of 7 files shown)


yaxu:
Excellent!

ndr-brt:
Ok, I noticed that the branch was not rebased to main, I updated it

yaxu:
Thanks :)

yaxu:
So @ndr-brt how does this work, where do the binaries end up? Are we able to trigger this without doing a release, for testing things out?

ndr-brt:
If I remember well, the `release` action needs a tag to be executed, so it creates a release with the binaries in the "releases" section. Maybe @polymorphicengine can correct me if I'm wrong

ndr-brt:
To test maybe we could make that run at every push, not only when there's a tag, then set the release step to run only when there's a tag

polymorphicengine:
yes that's correct. To try it out just change it to 
```
on:
  push:
    branches: [ main ]
```
I don't know how to conditionally run a step when there is a tag from the top of my head but it should be possible.

PR #857, opened 2021-10-25
Use better show instances for notes & rationals

mindofmatthew:
In the discussion from #807 it came up that the `Show` instance for notes is a bit confusing, because it exposes the structure of the Note type. I added a new show function that prints the numeric value followed by the note name. Here's what it looks like in the context of a pattern:

```
> n "c d6 20.75 -30"
(0>¼)|n: 0.0n (c5)
(¼>½)|n: 14.0n (d6)
(½>¾)|n: 20.75n (a6)
(¾>1)|n: -30.0n (fs2)
```

I also updated the `Show` instance for patterns to use the prettier rational representation (used by time spans) for rational values.

### DIFF
--- a/src/Sound/Tidal/Pattern.hs
+++ b/src/Sound/Tidal/Pattern.hs
@@ -647,8 +647,16 @@ instance NFData Value
 type ValueMap = Map.Map String Value
 
 -- | Note is Double, but with a different parser
-newtype Note = Note { unNote :: Double } deriving (Typeable, Data, Generic, Eq, Ord, Show, Enum, Num, Fractional, Floating, Real)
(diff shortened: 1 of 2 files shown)


yaxu:
Aside from the stack issue (which isn't particular to this PR), is this ready for merge?

mindofmatthew:
Yes, unless you (or anyone else) wants a different display for non-integer notes. I think the most accurate representation would be something like `2.7n (ds5 - 30 cents)` but that's maybe too long? Happy to drop the pitch names from non-integers or add a symbol like `+d5` or `<ds5` to indicate when a number is in between pitches.

yaxu:
Lets go with this for now, thanks!

Issue #853, opened 2021-10-17
Frequent Tidal crashes using jumpIn' code


Ouyuegu:
I had a series of mysterious Tidal crashes that, after months of trying various things out, I managed to whittle down to the `jumpIn'` command.

For example, when evaluating code like the following:

```
do
  jumpIn' "drill" 0 silence
  jumpIn' "beat" 0 silence
  jumpIn' "build" 0 silence
  jumpIn' "basssynth" 0 silence
  jumpIn' "basssynth2" 0 silence
  jumpIn' "stabs" 0 silence
  jumpIn' "giggles" 0 silence
  jumpIn' "brass" 0 $ (0.5 <~) $ loopAt 4 $ s "wg" # gain 1.1 # lpf 3000 # end 0.175
  jumpIn' "blues" 0 $ (0.05 ~>) $ slow 4 $ stack [ s "wg:2" # pan 0.2 # gain 1.1, s "wg:3" # pan 0.8 # gain 0.9 ] # lpf 7000
```

...Tidal would _sometimes_, but not _every_ time, crash. It would not spit out any error messages, and CPU usage was not too high nor too low. Upon Rebooting Tidal, it would inform me that it had closed due to the following exception:

`t> *** Exception: ghc: signal: 15`

After replacing all of these `jumpIn' "XXX" 0` commands with `jumpMod "XXX" 1` commands, the error stopped occurring.

I am running: Mac OS 11.4, Supercollider 3.12, Tidal 3.16.17.

Ouyuegu:
Hi everyone- I know this is closed, but I'm wondering at what point this fix will be included in a Tidal update... Or if there's something else I can do to implement it in the meantime?

Thanks!

PR #858, opened 2021-10-28
Switch to non-blocking pMap updates for transitions, which should fix…

bgold-cosmos:
… #853


Ouyuegu:
@bgold-cosmos  Hi there- I raised the original issue a while ago, and I see that the original issue was closed... I'm a bit confused as to what I would have to do now to affect this solution? I'm quite new to all this, so forgive the ignorance.

Thanks in advance!

ndr-brt:
I think you just need to upgrade tidal to the latest version: https://tidalcycles.org/docs/getting-started/upgrading/

PR #860, opened 2021-11-02
Fix mod

bgold-cosmos:
Add a `Moddable` typeclass so that `|%|` operators work with both "bare" numeric Patterns and ControlPatterns.


yaxu:
Nice, thanks!

PR #861, opened 2021-11-02
Fix tidal-listener install process

ndr-brt:
After last merge the build started failing: https://github.com/tidalcycles/Tidal/runs/4074323079


ndr-brt:
I tried also to fix the failing stack build but no success. 
Anyway I resolved some annoying warning

yaxu:
Great thanks !

PR #866, opened 2021-11-07
tidal-listener: Separate build from release

ndr-brt:
I separated the build job from the release job, and the latter would be executed only on a tag.
The artifacts will be available in the action section of the action workflow for 90 days (according to the github actions documentation)


ndr-brt:
e.g. the artifact for the linux build is available here (bottom of the page): https://github.com/tidalcycles/Tidal/actions/runs/1432003660

yaxu:
Thanks!

PR #876, opened 2021-12-08
Derive RealFrac for Note

### LINKED ISSUE #874: No instance for (RealFrac Note) arising from a use of ‘quantise’
[copied from <https://club.tidalcycles.org/t/quantise-type/3684>]

The quantise example fails because of a type error:
```
d1 $ s "superchip*8" # n (quantise 1 $ range (-10) (10) $ slow 8 $ cosine)
# release (quantise 5 $ slow 8 $ sine + 0.1)

No instance for (RealFrac Note) arising from a use of ‘quantise’
```

RealFrac is the type class that holds the round method that is used in quantize. Note is currently not an instance of RealFrac. Perhaps it should be? Especially in light of the comment (Note is Double)
```
class (Real a, Fractional a) => RealFrac a where
  round :: Integral b => a -> b

quantise :: (Functor f, RealFrac b) => b -> f b -> f b
quantise n = fmap ((/n) . (fromIntegral :: RealFrac b => Int -> b) . round . (*n))
```
one solution is to derive `instance RealFrac Note`, by extending this declaration
```
-- | Note is Double, but with a different parser
newtype Note = Note { unNote :: Double } 
  deriving (Typeable, Data, Generic, Eq, Ord, Enum, Num, 
    Fractional, Floating, Real)
```

ndr-brt:
Fix #874

### DIFF
--- a/src/Sound/Tidal/Pattern.hs
+++ b/src/Sound/Tidal/Pattern.hs
@@ -669,7 +669,9 @@ instance NFData Value
 type ValueMap = Map.Map String Value
 
 -- | Note is Double, but with a different parser
-newtype Note = Note { unNote :: Double } deriving (Typeable, Data, Generic, Eq, Ord, Enum, Num, Fractional, Floating, Real)
+newtype Note = Note { unNote :: Double }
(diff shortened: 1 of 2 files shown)


ndr-brt:
Don't know why the build-listener fails on windows, I don't think it depends by my commit

PR #880, opened 2021-12-27
Import safe-tidal-cli ghci simulator as tidal binary

yaxu:
From the mighty https://github.com/jwaldmann/safe-tidal-cli

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ out/
 .idea/
 *.lock
 *.iml
+*.swp
 
 # text editor temp files
 *~
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -1,0 +1,112 @@
+{-# language PatternSignatures, LambdaCase #-}
+
+import qualified Language.Haskell.Interpreter as I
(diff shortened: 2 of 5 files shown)


yaxu:
@jwaldmann How do you feel about this? Hope you don't mind, I thought safe-tidal-cli would make an excellent default tidal 'binary' and am giving it a go here.. I just adjusted the code block detection, so it can be a drop-in ghci replacement.

jwaldmann:
Yes fine. I'm glad you find this useful. 

On code block detection: "ghci-like" (what you implemented) means that a code block is
* a single line (so, delimited by `\n`)
* or several lines in `:{' .. ':}` ?

That's incompatible with "for flok" (what I had) where a code block is
* several lines, delimited by a blank line (so, `\n\n`)

Too bad that's incompatible. Perhaps make it switchable via command line option?

yaxu:
Great thanks!

Hm, I believe flok does wrap with `:{` `:}`:

https://github.com/munshkr/flok/blob/main/packages/repl/src/repl/tidal.ts#L16

But no problem to add the option if it's useful.

PR #881, opened 2021-12-27
Adds the jumpMod' transition

th-four:
Similar to jumpMod', with an extra argument: jumpMod' 1 n p will jump at the
boundary of the cycle p in a block of n cycles, where jumpMod would jump at the
limit of the n-cycle block.


yaxu:
Thanks !

PR #887, opened 2022-01-03
Hide contexts from Events per default

polymorphicengine:
https://club.tidalcycles.org/t/why-does-the-context-of-a-pattern-depend-on-whether-i-use-mini-notation-to-define-it/3767/4

In the above thread we discussed that it can be a little bit confusing that Contexts are in the Show instance for Events, but not for Patterns. 
I removed it, but there are still the functions `showFull` and `showEventFull` that show the Context of patterns/events.


yaxu:
Thanks!

PR #888, opened 2022-01-03
Define the Eq and Ord instances for the Event type by hand,

JeffreyBenjaminBrown:
so that otherwise-equal Events are considered equal
  even if their Contexts (debugging information) differ.


JeffreyBenjaminBrown:
For `Eq` I feel confident that this is the most natural implementation. For `Ord` there's another possibility: We could consider `Context`s when comparing `Event`s but just make that the last, rather than the first, thing they are compared on. It's a one-line change:
```
-relevant e = (whole e, part e, value e)
+relevant e = (whole e, part e, value e, context e)
```

PR #890, opened 2022-01-03
Define the Eq and Ord instances for the Event type by hand.

JeffreyBenjaminBrown:
This way, otherwise-equal `Event`s are considered equal
  even if their `Context`s (debugging information) differ.
  Also, `Event`s are compared first on when they happen
  (the `part` field), second on their `value`,
  and only last on their `whole` field (which is often `Nothing`).
  Sorting a `List` of `Event`s will thus cause the early ones
  to be first in the `List`, and contemporaneous ones
  to be adjacent in the list even if their `whole`s differ.

### DIFF
--- a/src/Sound/Tidal/Pattern.hs
+++ b/src/Sound/Tidal/Pattern.hs
@@ -555,7 +555,16 @@ data EventF a b = Event
   , whole :: Maybe a
   , part :: a
   , value :: b
-  } deriving (Eq, Ord, Functor, Generic)
+  } deriving (Functor, Generic)
+
+instance (Eq a, Eq b) => Eq (EventF a b) where
+  (==) x y = let relevant e = (whole e, part e, value e)
+             in relevant x == relevant y
+
+instance (Ord a, Ord b) => Ord (EventF a b) where
+  (<=) x y = let relevant e = (whole e, part e, value e)
+             in relevant x <= relevant y
+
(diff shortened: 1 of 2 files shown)


jwaldmann:
I feel a bit uneasy about this. The other suggested change (remove Context from instance Show Event) is just for human-readable output, but this here concerns `==` (etc.) that may be important for code. Are we sure that actual (derived) equality is never used (now, and in the future)?

The main motivation is for writing tests? There are are already some classes/functions for
comparing (sets of) events, defined in https://github.com/tidalcycles/Tidal/blob/669fde100a04ed1f86460a2f38ca618aa8d12150/test/TestUtils.hs#L34 (TolerantEq, compareP, comparePD). How does this relate?

If there are some Events where Context is important, and some where it is not, then
the True Haskell Way is to make them distinct types, so they cannot be confused,
and each can have their specific class instance(s).

JeffreyBenjaminBrown:
`stripContext` is certainly a handy thing to be able to do.

I wouldn't be terribly bothered if the changes to `Eq` didn't make it through. In fact even keeping `context` part of the `Ord` instance could be acceptable -- if it were considered last among fields, rather than first (as it is now).

yaxu:
Tidying up PRs, it looks like the show instance doesn't show context. Happy to look at a new PR that deals with alternative ways of comparing events, if there is a use case.

PR #895, opened 2022-02-14
tidal-listener: Add minimal install notes

gamar3is:
I was able to piece out this information from a couple of Github issues, and maybe the README's the place for that?


yaxu:
Thanks!

PR #898, opened 2022-02-16
Use ableton link for scheduling

Zalastax:
Continues from #891


yaxu:
Is this ready to merge now? As far as I can tell the automated build failures are down to a weirdo cabal path change. Hopefully an easy fix and can be done post merge.

Zalastax:
Hi @yaxu! This is fairly ready. The core of the Haskell code should be good now. However, if you look at the last commit you'll notice that I just now managed to (hopefully) squash a pretty huge bug. I therefore believe it would be best if someone else can try out some tricky cases before we merge, but it's up to you. We can also opt to squash the bugs after merging.

I have removed some functionality along the way. The getcps / getnow functions for example. Is it important to add them back before we merge? What about the destroyed functionality in tidal-listener?

OSX has not been tested. I'm fairly certain we need to change the compilation flags and tell link we're building on OSX. I can add this soon but I can't test that it actually works. An OSX user should preferably test before merging.

Finally, https://gitlab.haskell.org/ghc/ghc/-/issues/20918 has forced me to use a shared library on Windows. This will be problematic in that users might need to edit their PATH and the ghci command will be different. I'd prefer if we can find what causes this and fix it so that shared library is not used on any platform.

Zalastax:
As a reference for anyone who wants to test this, I simply use `cabal v1-install --force-reinstalls . ./tidal-link/` to build. On Windows I have also ensured that the shared library that gets built is on PATH, which for me means `C:\cabal\bin`. Other platforms don't use a shared library.

Zalastax:
I have prepared a branch where Windows does not use a DLL: https://github.com/Zalastax/Tidal/tree/tidal-link-no-dll. It compiles, but crashes in runtime. Windows users, please try to reproduce the issue reported in https://gitlab.haskell.org/ghc/ghc/-/issues/20918 and report your findings!

Zalastax:
Hi @yaxu,

This is ready for merging in my view.
It seems like the c++ options didn't work well in some CI-targets, so I removed the failing targets - I hope that's an OK resolution.

yaxu:
Hey it seems like the listener binary still isn't building for windows, is that right?

It would be best to try to support older ghc versions. So I think best add the ci checks back even if we merge them failing.

yaxu:
This is amazing work by the way, thanks very much! Looking forward to trying it out. I'd generally favour merge now and do any further dev later but it would be good to get the listener binaries building first

Zalastax:
Thanks for the kind words @yaxu!

Let me see what I can do. These are the things I know or will follow up on:

1. The tidal-link library requires GHC 9.2.4 on Windows.
2. hint does not yet support GHC 9.2.4.
3. tidal-listener uses hint.
4. cxx-options are not handled correctly in Cabal without https://github.com/haskell/cabal/pull/7072.
    * [ ]  I need to check what versions of Cabal it's included in and update CI to use those Cabal versions.
5. Executable version might work on Windows with GHC older than 9.2.4.
    * [ ] I should check if the binaries work on Windows with old versions of GHC and update the cabal files accordingly.

Zalastax:
Hi @yaxu, the CI is now passing. Can I squash and merge?

yaxu:
Yeah go for it!

PR #900, opened 2022-02-19
Add parsers for ints and floats that don't consume trailing whitespace

### LINKED ISSUE #897: parsing boolean patterns with consecutive *
https://club.tidalcycles.org/t/struct-behavior-in-mininotation/3886/2
It seems there is a bug in parsing boolean patterns that have a consecutive application of the `*` mini-syntax operator.

Examples:

`struct "t*1 t*1" $ s "bd"`
`while "t*1 t*1" rev $ s "bd"`

throw the following error:

```
Syntax error in sequence:
  "t*1 t*1"
        ^  
unexpected '*'
expecting rest, "[", "{", "<", "^", ".", "?" or end of input
```
just one application, or alternating applications work: `"t*1"`, `"t*1 t t*2"`.

if we replace the `t` by `1` it also works.

polymorphicengine:
Fixes #897 
The problem was that the built in `integer` and `float` parsers consume trailing whitespace, and thus seperating different expressions by whitespace was not possible.

### DIFF
--- a/src/Sound/Tidal/ParseBP.hs
+++ b/src/Sound/Tidal/ParseBP.hs
@@ -560,9 +560,9 @@ pRational :: MyParser (TPat Rational)
 pRational = wrapPos $ TPat_Atom Nothing <$> pRatio
 
 pRatio :: MyParser Rational
-pRatio = do 
+pRatio = do
   s <- sign
-  r <- do n <- try intOrFloat
+  r <- do n <- try (try pFloat <|> pInteger)
           v <- pFraction n <|> return (toRational n)
           r <- pRatioChar <|> return 1
           return (v * r)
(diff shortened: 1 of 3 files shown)


ndr-brt:
Great! Can you add a test to demonstrate this case working? So there won't be regressions in thr future

polymorphicengine:
Ah yes, good idea. Done!

src/Sound/Tidal/ParseBP.hs
```diff
@@ -560,22 +560,35 @@ pRational :: MyParser (TPat Rational)
 pRational = wrapPos $ TPat_Atom Nothing <$> pRatio
 
 pRatio :: MyParser Rational
-pRatio = do 
+pRatio = do
   s <- sign
-  r <- do n <- try intOrFloat
+  r <- do n <- try (try pFloat <|> pInteger)
```
ndr-brt, line 565: At this point `try pFloat <|> pInteger` could be the implementation of `intOrFloat`.
If so, everything will work as expected, except a test: 
https://github.com/tidalcycles/Tidal/blob/12c40a5b6d9fafcd3b895d9b90de1ac97483cbbf/test/Sound/Tidal/UITest.hs#L139

The correct event is:
`Event (Context [((1,1),(4,1))]) Nothing (Arc 0 0.5) (6 :: Int), Event (Context [((4,1),(5,1))]) Nothing (Arc 0.5 1) (0 :: Int)`

This will avoid duplication.

polymorphicengine: Yes you're right, we should probably replace `inOrFloat`. 
I don't really understand why this test fails, and I also don't see the difference between the events in the test file and the events you posted ? What am I missing?

ndr-brt: The test is wrong, it expect the context of the event to be from the character 1 to 4 of the pattern, but in reality is 1 to 3.
Whops, I pasted the wrong event! this is the right one:
`Event (Context [((1,1),(3,1))]) Nothing (Arc 0 0.5) (6 :: Int), Event (Context [((4,1),(5,1))]) Nothing (Arc 0.5 1) (0 :: Int)`

polymorphicengine: Ah, makes sense, thanks! Just added a new commit.

ndr-brt (approved):

yaxu:
Thanks both !

PR #904, opened 2022-04-06
Introduce `echo` and `echoWith`, deprecate `stut` and `stutWith`

### LINKED ISSUE #902: stut/stutWith inconsistency
As notified by some people (like @cleary and @felixroos), there's an inconsistency between `stut` and `stutWith`, as:

> stut is (times, feedback, time)
> stutWith is (times, time, func) 
> probably stut (times, time, feedback) would be more logical

I also point out that `stut` can be considered an ableist term (as it was done on supercollider, and deprecated: https://github.com/supercollider/supercollider/issues/5383)

So one option could be to deprecate `stut` and `stutWith` and define a new pair of functions with the right definition, the other would be to fix the inconsistency of `stut` and release a major version (probably including other breaking changes as well).

I would prefer to define a new pair of function and leave `stut`/`stutWith` there, mark them as deprecated and remove them in the next major release.
Some terms proposed for this rename are:
- `echo` (already used)
- `dup` (as done in supercollider)
- `comb`
- `ditto`
- `retrig`
- `rep`

ndr-brt:
Fix #902

* Remove `echo`, `triple`, `quad`, `double` aliases as seems like they are not used a lot (nor documented)
* Add `echo` and `echoWith` that have a more consistent signature than `stut` and `stutWith`, that will be removed.
* The `echo` function, unlike `stut`, applies only a simple gain multiplication on every step, as reported in the documentation.

Warning: this PR contains breaking changes

### DIFF
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE FlexibleInstances, OverloadedStrings, FlexibleContexts, BangPatterns #-}
 
 module Sound.Tidal.Control where
-
 {-
     Control.hs - Functions which concern control patterns, which are
     patterns of hashmaps, used for synth control values.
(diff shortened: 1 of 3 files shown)


ndr-brt:
@mxmxyz @felixroos I can't add you as reviewers, but since you participated to the discussion, your feedback is welcome

PR #910, opened 2022-05-02
Fix echo

ndr-brt:
As reported on discord, the first version of `echo` introduced in 1.8.0 is not working correctly, this PR fixes it.

As an addiction, made also `comparePD` print out the detail on failure.

@yaxu at this point, is `compareDefrag` still useful? Seemed to be used only for testing purposes, but at this point it's not used by anyone

@cleary please take a look at this

### DIFF
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -350,21 +350,21 @@ smash' n xs p = slowcat $ map (`slow` p') xs
 
     This adds a bit of echo:
     @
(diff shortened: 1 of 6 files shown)


yaxu:
I guess `compareDefrag` might be useful again someday.. But feel free to delete it if you don't think so.

yaxu (approved):

ndr-brt:
Ok, I deleted it. Now this is ready for merge

yaxu:
Thanks!

PR #916, opened 2022-05-28
if first argument of euclid is negative, apply euclidInv

### LINKED ISSUE #915: Mini-notation for euclidInv
What if inserting negative values into the first argument of euclid mininotation resulted in the inverted version of that euclidian rhythm. So that, for example:
```
"bd(-3,8)"
```
was the same as
```
euclidInv 3 8 $ s "bd"
```

polymorphicengine:
this is also patternable.
+ tests
fixes #915

### DIFF
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -718,6 +718,8 @@ In the above, three sounds are picked from the pattern on the right according
 to the structure given by the `e 3 8`. It ends up picking two `bd` sounds, a
 `cp` and missing the `sn` entirely.
 
+A negative first argument provides the inverse of the euclidean pattern.
+
 These types of sequences use "Bjorklund's algorithm", which wasn't made for
 music but for an application in nuclear physics, which is exciting. More
 exciting still is that it is very similar in structure to the one of the first
@@ -756,7 +758,7 @@ euclid = tParam2 _euclid
 
 _euclid :: Int -> Int -> Pattern a -> Pattern a
 _euclid n k a | n >= 0 = fastcat $ fmap (bool silence a) $ bjorklund (n,k)
-              | otherwise = _euclidInv (-n) k a
+              | otherwise = fastcat $ fmap (bool a silence) $ bjorklund (-n,k)
 
 {- | `euclidfull n k pa pb` stacks @e n k pa@ with @einv n k pb@ -}
 euclidFull :: Pattern Int -> Pattern Int -> Pattern a -> Pattern a -> Pattern a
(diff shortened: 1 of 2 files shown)


src/Sound/Tidal/UI.hs
```diff
@@ -809,7 +810,8 @@ euclidInv :: Pattern Int -> Pattern Int -> Pattern a -> Pattern a
 euclidInv = tParam2 _euclidInv
 
 _euclidInv :: Int -> Int -> Pattern a -> Pattern a
-_euclidInv n k a = fastcat $ fmap (bool a silence) $ bjorklund (n,k)
+_euclidInv n k a | n >= 0 = fastcat $ fmap (bool a silence) $ bjorklund (n,k)
```
ndr-brt, line 813: The implementation of this couldn't be:
```haskell
_euclidInv n k a = _euclid (-n) k a
```
?
it would avoid some duplication

ndr-brt (approved):
Great, thank you

src/Sound/Tidal/UI.hs
polymorphicengine: i think then we would have to replace the otherwise case in `_euclid` by the current definition of `_euclidInv`, no?

I agree that that might be a bit cleaner overall

ndr-brt: @polymorphicengine yes, probably at this point `euclidInv` could be also deprecated since it won't bring anything more than what `euclid` already bring

src/Sound/Tidal/UI.hs
```diff
@@ -756,7 +758,7 @@ euclid = tParam2 _euclid
 
 _euclid :: Int -> Int -> Pattern a -> Pattern a
 _euclid n k a | n >= 0 = fastcat $ fmap (bool silence a) $ bjorklund (n,k)
-              | otherwise = _euclidInv (-n) k a
+              | otherwise = fastcat $ fmap (bool a silence) $ bjorklund (-n,k)
```
ndr-brt, line 761: Is this correct? `bjorklund` should be called with `(n,k)`

polymorphicengine: I think it is correct, this is the case of `n < 0`, so `-n > 0` (otherwise bjorklund wouldn't produce anything). Note the reversed order of `a` and `silence`

ndr-brt: Yes, right, did a wrong assumption! So ok for me 👍

yaxu:
Thanks!

PR #918, opened 2022-05-31
Make chunk reverse direction with negative number

### LINKED ISSUE #917: Make `chunk` work as `chunk'` if called with a negative number
Currently, `chunk` is not working if called with negative numbers, would be great if:
```haskell
chunk "-3" "a b"
```

behave the same as
```haskell
chunk' 3 "a b"
```

ndr-brt:
Closes #917 

Further notes:
- removed `runWith` as it was deprecated since 2017
- deprecated `chunk'` `_chunk'` as their behavior is already covered by `chunk` and `_chunk`


yaxu:
Thanks!

PR #929, opened 2022-07-25
WIP sequences

yaxu:


yaxu:
> i think the typeclass definiton should rather be expecting a type of kind `* -> *` like `Pattern` or `Branch`

Agreed, fixed !

PR #931, opened 2022-07-27
overhaul of the chord parser

polymorphicengine:
to make this working, i had to change TPat to a generalised ADT to add a node for TPat_Chord that essentialy has three fields, one for each component of a chord: a `TPat a` where `a` is some Num type, a `TPat String` for chord names and a `[TPat [Modifier]]`  for modifiers. Due to some type trickery there is an additional field for making TPats instance of Functor (this is essentially the yoneda lemma applied, kind of like here https://stackoverflow.com/questions/17157579/functor-instance-for-a-gadt-with-type-constraint)

motivation for these changes was mainly this thread: https://club.tidalcycles.org/t/rfc-working-on-making-chord-naming-chordlist-more-consistent/2717/64

leaving the technical details behind, what these changes result in is a better seperation of syntax and semantics of chords, i.e., there is a corresponding semantic function `chord :: Num a => Pattern a -> Pattern String -> [Pattern [Modifier]] -> Pattern a`

further, every field in chords is patternable now and modifiers can be stacked
it should also be easy now to define new modifiers very quickly (like I did for the new drop modifier)

i also introduced two new functions `collect :: Pattern a -> Pattern [a]` and `uncollect :: Pattern [a] -> Pattern a` that collect/stack all events that have the exact same duration

### DIFF
--- a/src/Sound/Tidal/Chords.hs
+++ b/src/Sound/Tidal/Chords.hs
@@ -274,3 +274,44 @@ chordL p = (\name -> fromMaybe [] $ lookup name chordTable) <$> p
 chordList :: String
 chordList = unwords $ map fst (chordTable :: [(String, [Int])])
 
+data Modifier = Range Int | Drop Int | Invert | Open deriving Eq
+
+instance Show Modifier where
+  show (Range i) = "Range " ++ show i
+  show (Drop i) = "Drop " ++ show i
+  show Invert = "Invert"
+  show Open = "Open"
+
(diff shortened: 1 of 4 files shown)


yaxu (approved):

yaxu:
Looks good!

yaxu:
Ready for merge?

polymorphicengine:
Yes!

yaxu:
Thanks!

PR #933, opened 2022-08-03
a window-excursion for tidal.el

zzkt:
Starting tidal should now use the existing window configuration or split within an existing window. This seems like a more reasonable default (and can of course be configured as required outside tidal). There is also a prompt to restart the server if one is already running.


yaxu:
Great. Not sure what the random build error is about..

zzkt:
not sure why changing tidal.el would upset cabal? 🤷

PR #937, opened 2022-08-08
avoid race in loading bootscript during restart

zzkt:
In some cases BootTidal.hs could be loaded twice during the restart/start cycle. no longer.


yaxu:
ace

Issue #935, opened 2022-08-05
More random signals


fbous:
It would be nice to have better support for randomness in the standard library

Things I would like to see:

1. **More random distributions**: Support for the most common distribution functions (uniform, Gaussian, exponential, etc.) and easy methods to map between them.
2. **Correlated noise**: Add various instances of correlated noise
3. **Source of randomness in random functions**: Define variants of the functions that rely on randomness like `degrade` and `sometimes` that allow specifying which random function to use.

If this is a desired feature, I would be willing to implement the above points. In fact I'm implementing it for myself anyway but I think others could profit from this as well and having a discussion how to do it the *correct* way could also be fruitful.

I've noticed that #905 is still open so I could tackle this as well while I'm at it.

yaxu:
That would be great!

yaxu:
This could be a separate issue but I think there's some tidying to do around this.

e.g. there are a couple of functions like `chooseBy`, `select` and `degradeByUsing` that allow pattern signals other than `rand` to define how things are chosen or removed. The 'By' in `chooseBy` means which random pattern (or other signal) should be used. In `degradeByUsing`, that's what 'Using' means, and 'By' means by how much. `select` seems to be `chooseBy` but selects from a list of patterns rather than plain values.. It'd be good to standardise these names!

PR #942, opened 2022-08-14
WIP: Add random signals with common distributions

fbous:
Added a few random signals for #935.

It's a work in progress and I didn't get to properly test the different signals for performance and correct distribution yet.

Questions that I have so far:

1. Should the stuff be implemented in a separate file or in `UI.hs`? If the file `Random.hs` should be used, should I move the relevant functions from `UI.hs` there?
2. Not sure what the best method is to generate independent RV. Methods I see so far:
  - Scaling the time-seed by an irrational number
  - Shifting the time-seed by a large and/or irrational number
  - Applying Xorshift multiple times
  Right now I'm using the first but maybe the last idea could be better?
3. What other distributions would be useful / desired?

### DIFF
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -29,6 +29,7 @@ import Data.Ratio
 
 import Sound.Tidal.Pattern
 import Sound.Tidal.Core
+import Sound.Tidal.Random (_irand)
 import Sound.Tidal.UI
 import qualified Sound.Tidal.Params as P
 import Sound.Tidal.Utils
--- a/src/Sound/Tidal/Control.hs
+++ b/src/Sound/Tidal/Control.hs
@@ -29,7 +29,7 @@ import Data.Ratio
 
 import Sound.Tidal.Pattern
 import Sound.Tidal.Core
-import Sound.Tidal.Random (_irand)
+import Sound.Tidal.Random (irand)
 import Sound.Tidal.UI
 import qualified Sound.Tidal.Params as P
 import Sound.Tidal.Utils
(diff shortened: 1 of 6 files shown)


yaxu:
Fantastic!

1. I've probably vote for the separate file as-is, and move the existing random-related functions there. It's a bit tricky though - as `rand` is a signal, perhaps it should be in with the other signals like `sine`, `saw` etc, currently in `Core.hs`. A `Random.hs` is tidy enough though, and the difference between UI and Core is not at all well defined in any case..

I don't have useful insights for the other questions as I'm no expert on this stuff..

yaxu:
Is this ready to merge? It's kind of a breaking change but the next release will have other big changes so that's fine !

fbous:
No, it's still a work in progress

I would say that Point 1 of #935 is finished but I haven't addressed Point 2 and 3 yet. Since they are loosely related I guess we could have separate pull requests for each if that is desirable, e.g. if the next release is scheduled soon and we want to include the new random signals there.

However, it seems the checks are still failing for this PR so this definitely has to be fixed before merge. When is the new release scheduled?

yaxu:
The next release isn't scheduled at the moment, there's a bit of discussion about it here https://club.tidalcycles.org/t/releasing-tidal-1-9-0/4241

So no rush

yaxu:
Well 1.9.x is out now. But we can do a new release whenever, and it could be 1.10 as a breaking change.

yaxu:
A couple of years on, do you still have this in mind?
It would need retargetting for the `dev` branch.

fbous:
Thanks for checking in! I've tried to work on my PR a while ago but noticed that the target branch had evolved significantly and gave up again.

I think we can close this PR for now. If I ever get around working on Tidal again and it will still be relevant, I think it's easier to restart from a fresh branch and make a new PR. I'll keep the changes in my fork.

yaxu:
Apologies @fbous, a large development branch was abandoned, and a lot of reformatting has gone on on the new dev branch, but it should be much more stable from now, if/when you do get back to this.

sss-create:
@yaxu looking at this and the 'normal' function I backported in #1101, maybe we should skip adding normal and focus on implementing foundational algorithms presented here as building blocks? as @fbous said at least part 1 of #935 seems to be ready, so checking how to incorporate changes from this PR into main could be an idea. If @fbous has no time or desire to do so, I could check -- with an open outcome.

PR #946, opened 2022-08-26
Use c++ extra library for MacOS

giuseppelillo:
libstdc++ is deprecated on MacOS. Until we switch to `system-cxx-std-lib`, it is necessary to use `libc++` for MacOS builds.


Zalastax (approved):
Thanks!

PR #952, opened 2022-09-24
overhaul of the parser

polymorphicengine:
while i was working on the parser for chords i saw that the code of the parser was overall a bit messy and hard to understand
so i decided to clean the code up and make it generally more like a parser combinator, writing more basic parsers and combining them all together in the end
there were also too many parsers 'eating' whitespace.. i think it would be generally a bit better to be not so liberal about whitespace in mininotation anymore (for example allowing patterns like `"1  * 5 [4 5]   !!"`) 
i'm open to discussing this point a bit more though, since there might be examples where more whitespace will increase readability, as i implemented it now it is very strict (see the examples that fail in the test)
so this would actually be a breaking change 
it would be good to sort this out until the release of 2.0

while this is mostly just a refactoring / rewriting i think this is a good step towards new mini-notation features

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-16.31
+resolver: lts-20.5
 
 packages:
   - '.'
@@ -6,8 +6,7 @@ packages:
   - 'tidal-link'
 
 extra-deps:
-  - hosc-0.18.1
-  - microspec-0.2.1.3
+  - hosc-0.20
   - haskellish-0.3.2.1
 
-allow-newer: true
+


yaxu:
Thanks for the refactoring!

I don't really understand why the test examples would fail though
* `"{a !}%2"`
* `"<a!3 b ! c@5>"`
* `"<1!3 2 ! 3@5>"`

It seems to want whitespace after `!` etc, but I don't see why that should be a requirement?

polymorphicengine:
> It seems to want whitespace after `!` etc, but I don't see why that should be a requirement?

no, i think the error messages are misleading, it wants to eliminate the whitespace between the value and the operator

* `"{a!}%2"`
* `"<a!3 b! c@5>"`
* `"<1!3 2! 3@5>"`
should be fine

similar to this, there can't be any whitespace between a value and an operator except for `_`

yaxu:
I think whitespace is often nice to have, e.g. "b ! !" rather than "b!!" as it's a bit clearer that each exclamation point is a new event. What is the disadvantage of allowing whitespace there?

polymorphicengine:
I don't think in this case there is a real disadvantage, my thought on making it as strict as possible is so we can start thinking about cases like this where it can be useful and lead to better readability etc.

we can then document _the exact_ syntactic rules better and things like debugging the parser will become much easier and transparent

this is also why i'm for leaving this PR open until we discuss such things further.. we could maybe do a poll on the forum where people generally like or need whitespace

mindofmatthew:
In this case, it seems to suggest that there are two slightly different uses for the `!` operator (also `_` and `@`): one as a repetition of the previous value, where each `!` corresponds to an additional event. And secondly, as an operator in the expression of the form `x!n`. Combining the two meanings leads to some weird outcomes, for example: `"a! !2 !2"`, which produces three events when I'd expect it to produce five.

I'd propose something like `repeat = "!" | "!", integer` for the first case and `repeated value = value, "!", integer` for the second.

(As an aside, are `@` and `_` redundant, or do they behave differently?)

polymorphicengine:
> `"a ! !2 !2"`

hmm, i'm confused..
shouldn't it produce 8 ? if i were to use brackets: `"((a !) !2) !2" ~ "((a a) !2) !2" ~ "(a a a a) !2" ~ "a a a a a a a a"`

mindofmatthew:
It doesn't, because the `!` isn't really an operator that stacks like that. In one sense, it's more of a placeholder event: So `"a ! ! !" == "a a a a"`, not `"(((a !) !) !)"`

yaxu:
Yep, ! repeats the previous step, with !n as shorthand for multiple repeats.

`_` and `@` are aliases. They used to be `_` for elongating the previous step and `@t` for elongating by a factor of `t`, but now they both support both behaviours.

I think for ease of use, it's nice to allow whitespace anywhere that it wouldn't make an expression ambiguous.

JeffreyBenjaminBrown:
Hi! I just learned about this issue. Do you know about Text.Megaparsec.Expr? It's a great little module for defining a parser from a table of operators. You get to declare things prefix, postfix, infixr, infixl or infixn, and declare their precedences. As a result you don't have to anticipate how many and which operations will apply in what order.

polymorphicengine:
> Did you know about Text.Megaparsec.Expr?

I never really checked it out and just stuck with the thing i knew, but it sounds like it could be worth checking out!

PR #963, opened 2022-11-06
Add a Nix Flake for building and developing Tidal

mitchmindtree:
Hi folks! I've been digging into the Tidal code-base lately and thought I'd share the Nix flake I've been using to setup my Tidal environment and enable reproducible builds.

The following is from the flake's description:

> This adds a Nix flake for building and developing Tidal.
> 
> Packages are included for:
> - tidal
> - tidal-link
> - tidal-listener
> - tidal-parse
> 
> A `tidal-ghci` package is also included. This is a small script that starts an instance of `GHCi` with `Tidal` installed and with the `BootTidal.hs` file passed as the `-ghci-script`.
> 
> Packages can be built with `nix build .#tidal` or ran with `nix run .#tidal-ghci`.
> 
> A `devShell` is included that provides `cabal-install`, `stack` and all other build inputs for the tidal packages above included under a temporary shell. This shell can be entered with `nix develop`.

Note that [flakes](https://nixos.wiki/wiki/Flakes) are still an experimental feature of Nix! Their UX is such an improvement on the old workflow that I use them exclusively these days.

That said, I'd understand if the preference was to wait until the feature stabilizes in Nix before landing something like this.

Fwiw, I plan to keep using Tidal for the foreseeable future and am happy to help maintain this going forward (I'm also helping maintain [vim-tidal](https://github.com/tidalcycles/vim-tidal/graphs/contributors)).

A workflow has also been added that checks the formatting of the Nix code and checks that each of the packages build on x86 Linux and macOS.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,42 @@
+# A set of CI jobs for checking the Nix flake.
+
+name: "nix"
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  cancel-previous-runs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ github.token }}
+
+  nix-fmt-check:
+    needs: cancel-previous-runs
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


yaxu:
Thanks for this, I'm not familiar with nix so am not clear what this is, so if you're happy to continue maintaining it that's fine!

PR #967, opened 2022-11-12
Rename ghost'' to ghostWith

### LINKED ISSUE #958: rename `ghost''` to `ghostWith`
`ghostWith` is more inline with current function naming

> This variation of ghost adds another parameter to the ones present in ghost', which is the function used to modify the ghost notes. So here you can decide which changes will be applied to the ghost notes compared to the original notes.

https://tidalcycles.org/docs/reference/time/#ghost-2

### LINKED ISSUE #959: Make `ghost'` and `ghost''` parameters patternable
```
Type: ghost' :: Time -> Pattern ValueMap -> Pattern ValueMap
Type: ghost'' :: Time -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
```
Both functions return an error when trying to pattern the Time parameter - it'd be nice if this were possible :)

Zalastax:
Keeps ghost'' as an alias.
Closes #958

### DIFF
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -1728,11 +1728,15 @@ step' ss cs = fastcat $ map f cs
               | otherwise = silence
 
 
+-- ghost'' is kept for backwards compatibility
 ghost'' :: Time -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
-ghost'' a f p = superimpose (((a*2.5) `rotR`) . f) $ superimpose (((a*1.5) `rotR`) . f) p
+ghost'' = ghostWith
+
+ghostWith :: Time -> (Pattern a -> Pattern a) -> Pattern a -> Pattern a
+ghostWith a f p = superimpose (((a*2.5) `rotR`) . f) $ superimpose (((a*1.5) `rotR`) . f) p
 
 ghost' :: Time -> Pattern ValueMap -> Pattern ValueMap
-ghost' a p = ghost'' a ((|*| P.gain (pure 0.7)) . (|> P.end (pure 0.2)) . (|*| P.speed (pure 1.25))) p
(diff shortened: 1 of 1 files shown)


mindofmatthew:
Looking through the issues, I think this actually closes #958?

mindofmatthew:
I don't have approval permission, but I think this is worth merging. In general, I think v1.10 could benefit from cleaning up and standardizing the API a bit. Tidal currently exports [800 functions and 1300 params](https://docs.google.com/spreadsheets/d/1yVFXWt9Y5aDXxjR5iHIpgh7WSvT0YDj_48e2Z5QuIWg/edit?usp=sharing), only a fraction of which are documented. I'd propose the following guidelines for this process:

- Retain all old names as aliases for now
- Prefer `camelCase` version in function definitions and documentation (with `lowercase` aliases in code)
- Move away from "prime" function notation (`func'` or `func''`) (as it doesn't communicate anything about the function itself and it can't be represented in  Strudel)
- Adopt a more consistent vocabulary, eg:
  - `funcBy` generally means that an operation has an extra numeric argument, specifying how much or by what probability an operation is applied
  - `funcWith` generally means that an operation has an extra functional argument, specifying a function that further modifies the added/generated/modified pattern
 - Maintain parity with Strudel
 - Identify which functions are internal and explicitly exclude them from export lists

Anything I'm missing? I can do a first pass and drop a list of specific suggestions in an issue for folks to comment on, unless there's some other way we should approach this.

yaxu:
Ok merged!

@mindofmatthew thanks that's well thought through, I agree with every one of those points!

The complication is that I've put a lot of work into #947 on the cycseq branch, which is a major reorganisation towards Tidal version 2.0, based on the Tidal 'remake' (that went into vortex and strudel). Would you be up for looking at that instead? I haven't looked at it in a little while, it's not complete but it's passing most tests. It could be a good basis for standardisation.

mindofmatthew:
@yaxu Yes, I can definitely focus my attention on the `cycseq` branch, though I'd be inclined to submit parallel PRs against `main` if any particularly useful changes come up. I agree that 2.0 is the place for a broad reorganization.

PR #971, opened 2022-11-27
CHANGELOG gardening

mindofmatthew:
Looking back through, I noticed that some of the lists were displaying as blockquotes, and then I started fussing with inconsistent whitespace, and this is all a bit silly, but it's done


Zalastax:
Neat! And now tidy 😉

PR #972, opened 2022-11-27
Define `meta` and `replace`.

JeffreyBenjaminBrown:
`meta` lets you apply a pattern whose elements are pattern transformations like `fast 2`. `replace` makes it easier to use `meta`, and probably has other uses; it lets you turn a pattern of easy-to-write symbols (integers are what I usually use) into a pattern of anything else, by providing a lookup table.

  +meta :: forall a b.
  +        Pattern (Pattern a -> Pattern b) -> Pattern a -> Pattern b

  +replace :: forall a b. Ord a =>
  +           [(a,b)] -> Pattern a -> Pattern b

Here's a minimal example using both.

  This plays the pattern "ho*8"
  at normal speed for the first half,
  and at double speed for the second half.

  let slang = [ (0, id)
               , (1, fast 2) ]
     in p 1 $ meta (replace slang "0 1") "ho*8"

### DIFF
--- a/src/Sound/Tidal/Pattern.hs
+++ b/src/Sound/Tidal/Pattern.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
 {-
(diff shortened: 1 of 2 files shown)


JeffreyBenjaminBrown:
[On the forum](https://club.tidalcycles.org/t/has-anyone-defined-an-ap-like-function-f-f-a-f-b-f-a-f-b-where-f-pattern/3803/9), user @mvdirty managed to make one-liners out of both functions. His definition of `meta` is beautiful, and more general than mine:
```
meta :: Monad m => m (a -> m b) -> a -> m b
meta pf p = pf >>= ($ p)
```
It's simple enough that I no longer think `meta` deserves to exist.

His definition of `replace` is also beautiful, but complex enough that maybe it deserves to be defined in the codebase:
```
replace :: Show a => [(a, b)] -> Pattern String -> Pattern b
replace x = inhabit $ bimap show pure <$> x
```

The code I submitted is basically irrelevant at this point, but I'm leaving this pull request open just for discussion's sake.

yaxu:
Lets continue this interesting discussion here:
https://github.com/tidalcycles/Tidal/discussions/1150

PR #976, opened 2022-11-29
Comments for the `Event` type

JeffreyBenjaminBrown:
Add an explanation focusing on the interpretations of `part` and `whole`.

Based on this [discussion on the forum](https://club.tidalcycles.org/t/what-are-the-part-and-the-whole-in-an-event/4414/8).

### DIFF
--- a/src/Sound/Tidal/Pattern.hs
+++ b/src/Sound/Tidal/Pattern.hs
@@ -642,8 +642,33 @@ data Context = Context {contextPosition :: [((Int, Int), (Int, Int))]}
   deriving (Eq, Ord, Generic)
 instance NFData Context
 
--- | An event is a value that's active during a timespan. If a whole
--- is present, the part should be equal to or fit inside it.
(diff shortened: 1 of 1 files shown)


Zalastax (approved):

yaxu:
Thanks for this, but I think the explanation confuses a couple of things.

One use of 'whole' is to check whether it has an onset, and should therefore trigger a sound, it's true.

Another use is to find where the offset should be, i.e. if a sound should be triggered, how long it should be triggered for (if legato is on).

However, any kind of pattern can end up as a pattern of synth controls, including patterns of functions, numbers, strings etc. So they are important everywhere.

There are also uses that apply directly to any kind of pattern. For example when aligning events, `squeezing` cycles from one pattern into the events of another, with `squeezeJoin` and friends. 

Events without wholes are continuous events. It's not really that wholes aren't useful for those, it's simply that continuous events don't have wholes.

PR #977, opened 2022-12-04
Fix warning about type constraint

mindofmatthew:
Fixing a warning that keeps showing up when I build


Zalastax:
Great! Neatly solved 👍

PR #995, opened 2023-01-29
overhaul of the parser

polymorphicengine:
i kind of messed up when trying to update my last PR, sorry about that (still trying to figure out how to do this github thing)

it's the same code as before, so still needs some choices where to be more liberal with whitespace and where to be more strict.

as has been suggested i might also try out mega-parsec in the future

### DIFF
--- a/src/Sound/Tidal/ParseBP.hs
+++ b/src/Sound/Tidal/ParseBP.hs
@@ -25,12 +25,10 @@ module Sound.Tidal.ParseBP where
 
 import           Control.Applicative ()
 import qualified Control.Exception as E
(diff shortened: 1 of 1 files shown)


yaxu:
:)

yaxu:
Does this need redoing for the current dev branch?

polymorphicengine:
i don't think so, it was merged before

polymorphicengine:
i have had some plans to do an overhaul again, possibly switching to a parser generator. will take some time though

yaxu:
Ok I'll mark as done on #1076 for now

PR #996, opened 2023-01-31
Tidal listener update and rework

polymorphicengine:
I finally found the time to rework the tidal-listener with all the things i discovered while developing tidal-gui.

tidal-listener is now capable of interpreting any haskell statement (i.e. expressions of type `IO a`), in that way it is now very similar to ghci, trying to show the value resulting from the statement if possible.

further, `let`  and bindings of the form `x <- statement` are also valid statements that will bind variables that are kept in memory by the interpreter

for a full list of commands and commandline arguments see the README

### DIFF
--- a/tidal-listener/README.md
+++ b/tidal-listener/README.md
@@ -3,7 +3,7 @@ Experimental tidal OSC listener.
 
 ## Install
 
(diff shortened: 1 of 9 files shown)


yaxu:
Oh excellent !

PR #1000, opened 2023-03-02
Add open, inv and drop function

thgrund:
I would like to discuss these three new functions. I wanted to use inversions and drop chords independently from the chord notation in the mini notation and came up with this solution. 

My motivation comes from my system where I create harmonies and melodies in a totally different approach then we usually in TidalCycles: https://github.com/thgrund/mrreason-setup/blob/main/README.md 

But I think these functions could be useful in general.

### DIFF
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -28,7 +28,7 @@ import           Data.Bits (testBit, Bits, xor, shiftL, shiftR)
 
 import           Data.Ratio ((%), Ratio)
 import           Data.Fixed (mod')
-import           Data.List (sort, sortOn, findIndices, elemIndex, groupBy, transpose, intercalate, findIndex)
+import           Data.List (sort, sortBy, sortOn, findIndices, elemIndex, groupBy, transpose, intercalate, findIndex)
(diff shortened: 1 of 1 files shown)


thgrund:
And I am wondering why the git change looks so weird. Maybe I need to have a look at this again

thgrund:
And I realize that I need to change the name of the functions because of ambiguity

sss-create:
> And I realize that I need to change the name of the functions because of ambiguity

had a only quick look. what are the ambiguities?
n.b.: I didn't know there are no chord inversion functions -- great addition!
What else is there to do, except resolving conflicts?

thgrund:
Oh! I forgot, that this MR exists xD The ambiguity is, that there is already a `Prelude.drop` function and in TidalCycles the function `inv` already exists in `Sound.Tidal.UI`. 
And then we need to add unit tests for this. I already did this locally (I append theses tests to that comment). 

If you have the capacity to resolve the conflicts, and to rename the functions + add the unit test, then I would really appreciate it :-) 

Unfortunately I do have limited capacities right now.

[ChordsFuncTests.md](https://github.com/user-attachments/files/19044230/ChordsFuncTests.md)

PR #1006, opened 2023-04-19
Add capital-M aliases for major chords

heyarne:
I found it a bit confusing to see capital-M aliases missing when there are minor-m aliases for minor chords. From what I can tell it should be sufficient to add them to the `chordTable`.

I was not able to run the tests (Error: cabal: Cannot build the executables in the package tidal because it does not contain any executables) and I'm not familiar with haskell at all. From what I can tell there are no extra tests for the  that I would need to extend, is that correct? Let me know if there's anything else that would be good to add, or whether there's a reason maybe that these aliases have been left out.

Thanks!


yaxu:
Sorry for the delay in merging, looks good thanks!

Issue #954, opened 2022-09-29
Improve the default BootTidal.hs


yaxu:
It would be good to document it a bit more, switching from the old `startTidal` function to the more flexible `startStream` to make it easier for people to add new targets, etc (like in [this situation](https://club.tidalcycles.org/t/osc-out-of-tidal-ghc-network-error/4143/9)).

ejconlon:
> Can you see any drawbacks, e.g. in terms of reduced flexibility? Would this be a breaking change for some people?

I do not believe either of those are a concern. If people want to keep their existing boot file, it will still work.

sss-create:
closing this for now, because changes seem to be merged already and BootTidal.hs looks easier. :)

PR #1007, opened 2023-04-25
Add Sound.Tidal.Boot module with standard aliases

ejconlon:
Addresses #954 . This PR moves most of the content of the default BootTidal.hs file into the library module Sound.Tidal.Boot. This allows us to add Haddock documentation to the standard aliases that can be viewed on Hackage. These aliases can also be looked up on Hoogle. We also add few brief comments to the boot file.

The aliases in the original boot file were essentially partial applications of library functions with some in-scope tidal :: Stream value that is initialized at the top of the file. In this PR, we use a nullary typeclass class Tidally where tidal :: Stream with an orphan instance in the boot file instance Tidally where tidal = tidalInst to pass the reference to the stream to the alias functions, each of which has a Tidally constraint.

I have tested this PR manually. Please review for completeness of aliases and documentation, and maybe kick the tires yourself. I added an example.tidal file to the root of the repo for easy manual IDE integration testing, but please let me know if there is a better way.

### DIFF
--- a/BootTidal.hs
+++ b/BootTidal.hs
@@ -1,88 +1,43 @@
+:set -fno-warn-orphans
+:set -XMultiParamTypeClasses
 :set -XOverloadedStrings
 :set prompt ""
 
-import Sound.Tidal.Context
+default (Signal String, Integer, Double)
 
-import System.IO (hSetEncoding, stdout, utf8)
-hSetEncoding stdout utf8
+-- Import all the boot functions and aliases.
+import Sound.Tidal.Boot
 
+-- Create a Tidal Stream with the default settings.
+-- Use 'mkTidalWith' to customize these settings.
+tidalInst <- mkTidal
 
-let config = defaultConfig {cVerbose = True, cFrameTimespan = 1/20}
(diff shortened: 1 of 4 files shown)


yaxu:
Thanks! This is a bit delicate as editors have their own BootTidal.hs file that they use in some circumstances. I guess the old style BootTidal.hs will still work though?

yaxu:
Great, well the main branch is currently in a very 'experimental' state, so lets merge for now, and refine later if needed. Thanks a lot @ejconlon !

ejconlon:
@yaxu Please let me know if there is anything more I can do to refine `main` for release.

yaxu:
@ejconlon thanks, the big change now on main is moving things around to make room for a new 'sequence' type for beat-oriented representation that is dual to the current cycle-oriented one. I'm currently rewriting Sequence.hs type for clarity but hopefully will be ready for testing and polishing soon..

ejconlon:
@jwaldmann I am also surprised that it works! At first I tried implicit params, which seemed like a more "principled" approach that follows the usual scoping rules, but ghci did not support it well enough to use.

PR #1010, opened 2023-05-02
Remove tempo sync config options

mindofmatthew:
The OSC-based Tempo sync mechanism has been replaced with Ableton Link, so remove the unused sync options.


Zalastax:
I left them in to not break existing BootTidal.hs files / in case a new use case would appear. I agree that it's time to remove these parameters.

PR #1019, opened 2023-06-16
avoid loading stream module in other modules, so hint wont crash

polymorphicengine:
i explained this issue in more detail in the tidal discord under innards!


Zalastax (approved):

mindofmatthew:
Is it okay if we move it to `Sound.Tidal.StreamTypes` instead? I think that's a better place for it, since it's a stream-specific ID.

yaxu:
Thanks !

PR #1020, opened 2023-06-22
Remove tidal-parse from stack build to match cabal.project.

exitmouse:
`stack build` is broken on a new checkout because it attempts to build tidal-parse.


yaxu:
Thanks! tidal-parse will need to be fixed but not worth doing that until things are more stable.

PR #1025, opened 2023-06-29
change streamGetnow to factor in the processAhead and also destroy the sessionstate

polymorphicengine:


Zalastax:
What is the expected behavior here? To get the cycle that would right now be hitting the speakers or roughly get the cycle that we would be processing at this moment? Reading the old code it seems like the latter is what was used, which would correspond to the change you are proposing. But I want us to reflect on what is most useful. Do we have some clear use cases for this function? If it is most used for the `once` function, then adding `processAhead` seems somewhat wrong.

The old code:
https://github.com/tidalcycles/Tidal/blob/a05892257b5be001d1161b9b0849395f11cf5ef7/src/Sound/Tidal/Stream.hs#L714-L717

polymorphicengine:
> Do we have some clear use cases for this function?

The intended use is when someone wants to use the time of the stream to query a pattern for something (for example to highlight the pattern in the editor). When using this i encountered that with fast speeds there was a misalignment with the event that was playing and the one that was being highlighted. Adding the processAhead fixed the issue for me.

so maybe we should rename this function to something else and keep the old one as streamGetnow?

Zalastax:
Great idea!

What about `processGetAhead`?

polymorphicengine:
i named it `streamGetAhead` now :)

Zalastax (approved):
Good catch, that's better and probably what I actually meant 😸 Great job!

yaxu:
Is this ready to merge?

polymorphicengine:
i think so!

yaxu:
Thanks a lot!

Issue #982, opened 2022-12-23
Extendable OSC controller


Zalastax:
As requested by ghales in https://club.tidalcycles.org/t/rfc-improving-the-interface-for-interacting-with-tidal-streams/3528/4, Tidal could allow for extending the OSC controller. The requested idea is to allow callbacks to incoming OSC messages.

My suggested way to implement this would be to make the whole OSC controller replaceable and provided through configuration. The default configuration would be the current OSC controller. We should also create a function that returns an OSC controller that has the same base behaviour as currently + accepts a list of callbacks that are invoked for each incoming message.

mindofmatthew:
Sooo, as it happens, I've actually built a mostly-complete, if still speculative, version of that [in this branch](https://github.com/mindofmatthew/Tidal/tree/listener). There's still some bugs with it, and maybe it makes sense broken up into a couple of PRs, but I can go back in soon and polish it up, and then we can do some review.

Zalastax:
Great, @mindofmatthew! Some polish and we should be good to go 👍

Compared to my suggestion, this doesn't make the whole OSC controller replaceable. Do you agree that we should implement that in addition to the changes you are already introducing? My view is that it would be good flexibility to provide.

mindofmatthew:
I haven't added the functions yet, but this would allow you to attach individual controls to a stream, like so:

```haskell
-- Add a new listener action (the returned Int can be used to remove the listener action later)
streamAddListener :: Stream -> String -> ([Value] -> IO ()) -> IO Int

-- A BootTidal abstraction
onOSC :: String -> IO () -> IO Int
onOSC name action = streamAddListener tidal name (\_ -> action)

onOSCArgs :: String -> ([Value] -> IO ()) -> IO Int
onOSCArgs = streamAddListener tidal

-- Which could be used like
onOSC "/button_1" $ d1 $ s "bd sn"
onOSC "/button_2" $ d2 $ silence

onOSCArgs "/knob_1" (\v:_ -> all (# lpf v))
```

I could see the usefulness of swapping out the listener at a higher level, but I don't have a clear sense of how it would fit in with this yet. Can you give me an example of how you're imagining it would be used?

Zalastax:
I would like to make the OSC controller abstract to the Stream module and not tie it directly to the new Sound.Tidal.OSC.Listener module. The motivation for my suggestion is to make the core simpler. It could also let someone experiment with other ways of implementing OSC controller extensibility or completely replacing the base commands.

[...]

PR #1027, opened 2023-07-18
Split controller OSC off from SuperDirt handshake OSC

### LINKED ISSUE #793: ControlAddr should default to 0.0.0.0 if a target isn't localhost
Tidal now uses the control socket to send OSC *out* as well as receive OSC. This is so that superdirt knows where to send replies. However if superdirt is running on a remote host, this means that the default ControlAddr to listen on of localhost (127.0.0.1:6010) stops OSC from being sent.

The default should probably be to listen on all IPs, i.e. `0.0.0.0`, if there is a target other than 127.0.0.1. However people should be able to set a specific IP if they want.

Relatedly, the oHandshake setting on Target is ignored. If it's set to True, maybe OSC should be sent out on an arbitrary IP.

mindofmatthew:
This is an initial step towards disentangling the controller input logic from the `Stream` module (as discussed in #982). Specifically, it removes the SuperDirt handshake process from the controller thread and instead handles all that per-target using the individual SuperDirt sockets. This adds support for more unusual setups (e.g. one Tidal instance running multiple SuperDirt instances at multiple addresses), but should function identically in most cases.

Fixes #793—this uses totally separate addresses for SuperDirt communication and external controllers, so it's not an issue if one is on localhost and the other isn't.

(I've done some tests on my machine, and everything seems to be working the same as in 1.9.4. I started to write some unit tests, but couldn't figure out how to write tests that didn't actually send things over UDP and got nervous that using real UDP communication could result in flaky tests.)


mindofmatthew:
Pinging @Zalastax, who's expressed interest in OSC listener/stream logic, but happy to hear from anyone else with thoughts as well

src/Sound/Tidal/Stream.hs
```diff
@@ -252,35 +249,51 @@ startStream config oscmap
...
-sendBndl :: Bool -> (Maybe O.Udp) -> Cx -> O.Bundle -> IO ()
-sendBndl isBusMsg (Just listen) cx bndl = O.sendTo listen (O.Packet_Bundle bndl) (N.addrAddress addr)
+sendBndl :: Bool -> Cx -> O.Bundle -> IO ()
+sendBndl isBusMsg cx bndl = O.sendTo (cxUDP cx) (O.Packet_Bundle bndl) (N.addrAddress addr)
   where addr | isBusMsg && isJust (cxBusAddr cx) = fromJust $ cxBusAddr cx
              | otherwise = cxAddr cx
-sendBndl _ Nothing cx bndl = O.sendBundle (cxUDP cx) bndl
 
-resolve :: String -> String -> IO N.AddrInfo
+resolve :: String -> Int -> IO N.AddrInfo
```
Zalastax, line 294: Would it be a good idea to introduce a type synonyme here?
type Port = Int

src/Sound/Tidal/Stream.hs
```diff
@@ -213,28 +212,27 @@ startStream :: Config -> [(Target, [OSC])] -> IO Stream
...
        tidal_status_string >>= verbose config
        verbose config $ "Listening for external controls on " ++ cCtrlAddr config ++ ":" ++ show (cCtrlPort config)
        listen <- openListener config
 
-       cxs <- mapM (\(target, os) -> do remote_addr <- resolve (oAddress target) (show $ oPort target)
-                                        remote_bus_addr <- if isJust $ oBusPort target
-                                                           then Just <$> resolve (oAddress target) (show $ fromJust $ oBusPort target)
-                                                           else return Nothing
+       cxs <- mapM (\(target, os) -> do remote_addr <- resolve (oAddress target) (oPort target)
+                                        remote_bus_addr <- sequence (resolve (oAddress target) <$> (oBusPort target))
```
Zalastax, line 223: Smart! This uses the fact that Maybe is a Monad so sequence is the same thing as what was done here before.

Zalastax (approved):
Looks good to me!

yaxu:
Sorry for being a bit slow, somehow missed this, but looks great to me!

jarmitage:
@mindofmatthew which of the 4 points you mentioned in https://github.com/tidalcycles/Tidal/issues/982#issuecomment-1620345088 does this address? I'm guessing 1-2? Good stuff!

mindofmatthew:
While trying to verify that this works correctly, I discovered that _Tidal_ doesn't work correctly (see #1030), so I'm going to try and patch that here as well.

@jarmitage Yes, this basically cleans up the Stream module a bit. It doesn't totally separate the listener (which is my eventual plan), but it splits up two listener functions (external controllers vs. getting info from SuperDirt) that are currently combined.

yaxu:
@mindofmatthew are you happy to merge? 

(I'm splitting out stream into a separate tidal-stream package on the cycseq branch.)

mindofmatthew:
@yaxu Sure! I've tested and confirmed that this works. It does have a bug when SuperDirt has different control busses than 0-127 (#1030), but that's also present in 1.9, and I can address it in a separate PR.

yaxu:
Thanks a lot!

PR #1031, opened 2023-08-23
Rejig everything

yaxu:
A lot of refactoring (a lot still missing and broken)
Patterns now have metadata saying how they should be combined, which should be used by all patternified parameters
The main package is now split into tidal-core, tidal-stream and tidal-params-dirt (along with tidal-link)


yaxu:
Some cherry picking of ideas needed now 2.0-dev is likely abandoned.

PR #1036, opened 2023-09-13
fix minor7sharp9 chord (the 9th wasn't sharp)

cleary:
tbf, it's a redundant chord (the sharp 9 is the same as a minor 3rd already in the chord), but if people are using it then it should be left in place


yaxu:
Thanks!

PR #1042, opened 2023-10-11
add metatune param

ahihi:
adds support for the [`metatune` functionality](https://github.com/musikinformatik/SuperDirt/issues/273) recently merged into SuperDirt.

to summarize:

- WAV files can contain pitch metadata in a `smpl` chunk, commonly used by samplers for automatic pitch mapping
- SuperDirt now reads and stores this metadata for each loaded sample
- when the `dirt_sample` synth is passed `metatune = 1`, it tunes sample playback relative to the sample's pitch metadata

so given e.g. an instrument sample `foo:2` recorded at F#4 and tagged as such in its metadata, it is now possible to do `d1 $ note "c3" # s "foo:2" # metatune 1` and it will actually play at C3!

`metatune` defaults to `0`, in which case C4 is used as the reference pitch, just like before. C4 is also the fallback when `metatune = 1` but the specified sample lacks pitch metadata.

### DIFF
--- a/bin/generate-params.hs
+++ b/bin/generate-params.hs
@@ -178,6 +178,7 @@ genericParams = [
   ("f", "loop", "loops the sample (from `begin` to `end`) the specified number of times."),
   ("f", "lophat", ""),
   ("f", "lsnare", ""),
+  ("f", "metatune", "A pattern of numbers. Specifies whether the pitch of played samples should be tuned relative to their pitch metadata, if it exists. When set to 1, pitch metadata is applied. When set to 0, pitch metadata is ignored."),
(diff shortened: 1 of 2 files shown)


yaxu:
Thanks! I'm in the process of moving things around on the `cycseq` branch. I'll leave this PR open for now, and merge after #1031.

mindofmatthew:
@ahihi Thanks! Would you mind submitting a PR with this change against the [1.9-dev](https://github.com/tidalcycles/Tidal/tree/1.9-dev) branch? We're using the main branch for developing Tidal 2.0, but if you submit a patch on that branch it can get published sooner!

ahihi:
sure thing! closing this one.

Issue #1030, opened 2023-08-23
Tidal doesn't use control bus mappings


matthewkaney:
While testing #1027, I discovered that current Tidal (1.9.4) isn't correctly using the control busses allocated SuperDirt, but is instead just always using busses 0-127. This usually isn't a problem, since most users just run SuperDirt on its own and it gets allocated that block of busses, but if you allocate more busses than that (for example by running `SuperDirt.start` multiple times), then the control bus indices out of sync and don't work correctly.

To reproduce:
1. Evaluate the following code in SuperCollider to start a SuperDirt server. Because 20 separate control busses are allocated first, SuperDirt gets the next set of 128 busses (20-147).
    ```supercollider
    (
    s.reboot {
        s.options.numBuffers = 1024 * 256;
        s.options.memSize = 8192 * 32;
        s.waitForBoot {
            ~otherBusses = Bus.control(s, 20); // Allocate some control busses
            ~dirt = SuperDirt(2, s); // Set up superdirt
            ~dirt.start(57120, 0 ! 12);
            s.dumpOSC(1); // Turn on server OSC dumping for test
        };
    };
    )
    ```
2. Start up Tidal, and evaluate the following line:
    ```haskell
    import Control.Concurrent.MVar
    ```
3. Now, evaluate the following line in Tidal. It should print out an array starting with 20 (`[20, 21, 22, ...etc]`), demonstrating that Tidal received the correct bus mappings from SuperDirt.
    ```haskell
    readMVar (sBusses tidal)
    ```
4. Now, evaluate the following pattern in Tidal:
    ```haskell
    d1 $ s "supersquare" # n "c4" # lpfbus 0 "[80 8000]*4"
    ```

The _correct_ behavior would be to produce a pulsing tone because of the alternating filter cutoff. However, SuperDirt is listening for input on control bus 20 (the 0th bus allocated to SuperDirt), but Tidal is sending its control messages to control bus 0 on the server. Because of this, the tone should be barely audible as it has a default filter cutoff of 0hz (I believe). SuperCollider should also be dumping OSC messages to the console that confirm this:
<pre>
[ "#bundle", <i>timestamp</i>,
  [ "/c_set", <b>0</b>, <i>value</i> ]
]
</pre>

yaxu:
I believe this was fixed in #1051, thanks to @matthewkaney !

PR #1051, opened 2023-11-29
Split controller OSC off from SuperDirt handshake OSC (Redux for 1.9)

matthewkaney:
This cherry-picks the changes in #1027 into the `1.9-dev` branch, since they aren't specifically related to 2.0 and would be nice to include in 1.9.5 (the most practical benefit is that this removes the "Tidal can't connect to SuperDirt without a control port" error state that occasionally scares new users).

Marking this as a draft until I test this locally and verify that everything works. This duplicates an already-merged PR, so it should be fine, but it would be awesome if anyone wants to give it a quick look over to make sure there's nothing questionable happening.


mindofmatthew:
I'm holding this off until 1.10 so it can be incorporated into #1059

mindofmatthew:
Did some testing and it looks like this is all working on my end. To recap: SuperDirt handshakes/bus mappings are handled by individual targets, rather than by the stream. This means that each handshake occurs on a separate UDP port, and SuperDirt targets no longer use the Tidal control listener port for sending messages.

Fixes #1030 as well—a bug where the control bus mappings weren't actually being properly used to address bus messages (and therefore bus params wouldn't work if SuperDirt allocated a different block control buses than 0-n)

In general, this is part of a broader effort to separate out different pieces of stream functionality (ie, consolidate the listener logic, consolidate SuperDirt-specific features, etc). I know @polymorphicengine has been thinking along these lines, so let me know if you have any thoughts!

polymorphicengine:
i think this is looking good!
are you also planning to move into the direction of the extendable OSC controller? (https://github.com/tidalcycles/Tidal/issues/982)

i mean it is actually pretty much there, we only need to abstract over the act function inside ctrlResponder so people can extend it like

```
act' :: O.Message -> IO ()
act' matchingMessage = customAction
act' m = defaultAct m
```

mindofmatthew:
> are you also planning to move into the direction of the extendable OSC controller?

Yes, though I think in a separate PR. My (maybe controversial) thought is that the Stream object shouldn't own the controller/listener at all, since the listener is just a wrapper around functions that the Stream object exposes. Or else we move to a point where all of the relevant logic is split across a bunch of small logical components (a clock, a listener, some targets, a playmap abstraction) and `Stream` is just a set of MVars that holds together a default (but entirely optional) Tidal setup.

sss-create:
uh nice, @matthewkaney!

yaxu:
Ready for merge?

ahihi:
i did some testing on this just now, it looks good to me and solves my control bus woes 😍

matthewkaney:
Oh great! I wasn't sure if everything still worked after the messy merge, but that's good to hear.

I'll test a bit this evening to see if I catch anything amiss in how it behaves and look into why the builds are failing. After that, this should be good to merge tomorrow morning!

matthewkaney:
@yaxu Yes, I think that can be merged! I guess the latest build issue is some error with the Ormolu action, possibly related to the fact that this PR is from a fork rather than a branch on the main repository? Good to address at some point, but doesn't seem like it will be an issue here after merge

yaxu:
Excellent thanks! I think the build issue was a small formatting thing that was already automatically fixed.

Issue #1054, opened 2023-12-05
`tidal-listener` builds in Nix are failing on 1.9


mindofmatthew:
Currently, the Nix build of `tidal-listener` is failing on [1.9-dev](https://github.com/tidalcycles/Tidal/tree/1.9-dev) with a pretty cryptic error message. My vague sense from searching around is that this means GHC is not loading some internal interfaces in the correct place? We disabled `tidal-listener` builds in `main` because of potential 2.0 incompatibilities, but I don't think that's the issue here—in particular, the listener seems to build fine outside of nix.

[Here's an example the error in question](https://github.com/tidalcycles/Tidal/actions/runs/7093750696/job/19307755096?pr=1051)

I know that the listener needs to be compiled with the `--dynamic` flag (or at least does on Windows?) Perhaps some nuance of the compilation setup isn't getting applied correctly in the nix build?

I don't know anything about nix, and only slightly understand the tidal-listener, so @mitchmindtree or @polymorphicengine, do either of you have any thoughts about what might be going on here? Thanks so much :sweat_smile:

mitchmindtree:
After a quick look I'm not too sure off the top of my head :thinking: 

Curiously, while `nix build .#tidal-listener` fails, doing `nix develop` and then `cd tidal-listener && cabal run` appears to work without issue. This does indicate there's some kind of impurity in the non-nix `cabal build` approach that we're not capturing in the `tidal-listener` flake package output.

Could you elaborate on the need for `--dymanic` you mentioned? Does cabal try to link to a system dynamic libs while building `tidal-listener` or something?

mitchmindtree:
I kept your `--dynamic` hint in mind and had a look at `tidal-listener.cabal` file.

It looks like commenting out these `ghc-options: -dynamic-too` lines in the `tidal-listener.cabal` file allows `nix build .#tidal-listener` to work without issue.

https://github.com/tidalcycles/Tidal/blob/5d20eecb303a2669d0f5d475f40c8cc7f679521e/tidal-listener/tidal-listener.cabal#L37-L38

The reason it was added appears to be something to do with `haskell-lint`? This issue is linked above:

https://github.com/haskell-hint/hint/issues/156

@polymorphicengine it looks like that issue has since been closed, but I'm unfamiliar with haskell-hint - are we able to remove the `-dynamic-too` ghc option? If so that would make this an easy fix :smile:

polymorphicengine:
[...]

PR #1062, opened 2024-01-16
More build fixes

mindofmatthew:
Unify conditions under which various build actions occur, and disable Nix build of tidal-listener because there's not a clear path forward there and the listener has always been experimental (see #1054).

### DIFF
--- a/.github/workflows/listener-build-linux.yml
+++ b/.github/workflows/listener-build-linux.yml
@@ -1,6 +1,11 @@
 name: build-listener-linux
-
-on: [push, pull_request]
+on:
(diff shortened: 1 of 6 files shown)


mindofmatthew:
All checks are passing! @yaxu, can you let me know if this is good with you whenever you have time?

Two build changes:
* `tidal-parse` dependency updates. Fixes #1069
* Disables automatic builds of `tidal-listener` on Nix, since it's not clear that anyone uses it, and hypothetical users would probably want (maybe need?) to build it locally anyway. Fixes #1054

PR #1063, opened 2024-01-19
Fix negative numbers not working for boolean euclids 

geikha:
As you may remember, with #915 and #916, we added the possibility of using negative values for the first argument of euclidian rhythms. However we forgot to replicate this behaviour for boolean euclidian patterns.

This fixes the following bug I've encountered:

```haskell
-- works
d1 $ mask "t(5,16)" $ s "808lt*16" |- note (saw*2) # shape 0.3 # octave 6

-- goes silent:
d1 $ mask "t(-5,16)" $ s "808lt*16" |- note (saw*2) # shape 0.3 # octave 6
```


mindofmatthew:
Looks good to me!

PR #1080, opened 2024-04-15
bump upper bounds of network dependency to 3.3

yaxu:
Seems fine from a quick local test.


yaxu:
I already updated this directly on hackage for 1.9.5 https://hackage.haskell.org/package/tidal-1.9.5

yaxu:
Are there any windows users about who could work out why the new network package version isn't building here?

PR #1081, opened 2024-04-17
Cherry pick new Boot changes and adjust to incorporate recent changes

### LINKED ISSUE #1074: BootTidal.hs breaking change + use of `defaultConfig`
I just ran into a bug we didn't catch with #1059: because it removes the `cFrameTimespan` from the config object, the default `BootTidal.hs` throws an error. This is pretty minor, but has a couple implications that are worth discussing:

### Is this breaking change okay?

Given that this has been in the default boot file for a long time, this is likely going to break everyone's custom BootTidal file. Personally, I think this is fine (breaking changes happen, and we can increment the version number and document migration steps accordingly), but I understand if anyone's hesitant. The backwards-compatible approach would be to include [all (or most) of the new Clock configuration options](https://github.com/tidalcycles/Tidal/blob/dev/tidal-link/src/hs/Sound/Tidal/Clock.hs#L48-L56) in the main `Config` datatype, and let `startStream` copy values individually.

That said, perhaps this is a good release to rework the boot file generally (see #954), incorporating the changes from #1007. We should also make sure that editors are falling back to the `BootFile.hs` installed with Tidal (e.g. tidalcycles/vscode-tidalcycles#40).

### Why don't we just use `defaultConfig` in the first place?

I'm also not totally sure why the default boot file updates the default config option anyway, given that the new values (`{cVerbose = True, cFrameTimespan = 1/20}`) are the same as the defaults. Is this desirable (perhaps so that people have example code to work from), or is it accidental or unnecessary?

@polymorphicengine and @yaxu probably have thoughts, but also interested in what others have to say!

mindofmatthew:
I went ahead and pulled together the changes from #1007 and then modified it to work with the current state of 1.10.

This fixes #1074, and generally pushes the "better BootTidal.hs" initiative forward (see #954) , but there might be other things we want to address in that regard.

### DIFF
--- a/BootTidal.hs
+++ b/BootTidal.hs
@@ -1,82 +1,43 @@
+:set -fno-warn-orphans
+:set -XMultiParamTypeClasses
 :set -XOverloadedStrings
 :set prompt ""
 
-import Sound.Tidal.Context
+default (Signal String, Integer, Double)
 
-import System.IO (hSetEncoding, stdout, utf8)
-hSetEncoding stdout utf8
+-- Import all the boot functions and aliases.
+import Sound.Tidal.Boot
 
-tidal <- startTidal (superdirtTarget {oLatency = 0.05, oAddress = "127.0.0.1", oPort = 57120}) (defaultConfig {cVerbose = True, cFrameTimespan = 1/20})
+-- Create a Tidal Stream with the default settings.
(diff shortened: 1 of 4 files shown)


yaxu:
Excellent thanks!

PR #1082, opened 2024-04-18
add muting with `_` for p and d1-16

geikha:
As [said on the Discord](https://discord.com/channels/779427371270275082/779737254465765426/1226743211121315920):

> An idea that could be interesting for performing:
> 
> ```hs
> _d1 _ = d1 silence
> _d2 _ = d2 silence
> -- ... etc
> _p x = p x $ silence
> ```
> 
> what if we define underscore functions for each default stream (d1, d2, d3, etc)?  this way one could easily silence or "mute" a pattern via adding an underscore and evaluating the same block of code instead of having to write `d1 $ silence` at the bottom and evaluating that

Wondering if this would have to be replicated on the pulsar default BootTidal files and such

---

Also see https://github.com/tidalcycles/strudel/pull/1059


yaxu:
Thanks, I've just merged #1007 though which means these aliases will need moving to `Sound.Tidal.Boot`.

That PR will reduce the need to update editor default BootTidal.hs files in the future.

yaxu:
Have you tested this? I'd have thought the functions would need adding to the export list at the top

geikha:
aghh my bad,, I _thought_ I tested this but I was using my own bootfile still 🤦‍♂️

src/Sound/Tidal/Boot.hs
```diff
@@ -107,6 +141,14 @@ only = (hush >>)
 p :: Tidally => ID -> ControlPattern -> IO ()
 p = streamReplace tidal
 
+-- | Silences a specific stream, regardless of ControlPattern input. Useful for rapid muting of streams
+_p :: Tidally => ID -> ControlPattern -> IO ()
+_p n _ = streamReplace tidal n silence
```
mindofmatthew, line 146: Might be good to use a parameter other than `n` here to avoid the warning? The Stream module functions use `k` (for pattern key) a lot in this situation

geikha: Sure! Just changed it

PR #1083, opened 2024-04-19
Add `getbpm` and `setbpm` to boot functions

### LINKED ISSUE #1018: add streamGetbpm to Stream.hs
`bpm <- Link.getTempo ss` is already available here, so it would be nice to add `streamGetbpm` here:

https://github.com/tidalcycles/Tidal/blob/bc03ed82496e1e84e7b814401af074006fc1fd82/src/Sound/Tidal/Stream.hs#L754-L759

mindofmatthew:
We have these functions in the Stream module now, so create aliases in `Boot.hs`

Fixes #1018

### DIFF
--- a/src/Sound/Tidal/Boot.hs
+++ b/src/Sound/Tidal/Boot.hs
@@ -27,6 +27,8 @@ module Sound.Tidal.Boot
   , setCycle
   , setcps
   , getcps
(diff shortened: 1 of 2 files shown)


yaxu:
Thanks!

Issue #979, opened 2022-12-21
`slice` of 1 doesn't work


mindofmatthew:
`slice "1" "0" $ s "bd"` produces the following pattern `(0>1)|begin: 0.0f, end: 0.0f, s: "bd"`. This doesn't seem to be correct—a slice of 1 should produce the full sample.

Incidentally, `randslice` behaves correctly: `randslice "1" $ s "bd"` produces `(0>1)|begin: 0.0f, end: 1.0f, s: "bd"`

grddavies:
I took a look at this today and came up with two possible fixes:

I didn't like this as much as it handles n = 1 as a special case, but it 'does less':

```haskell
slice' :: Pattern Int -> Pattern Int -> ControlPattern -> ControlPattern
slice' pN pI p = P.begin b |> P.end e |> p
  where b = div' <$> pI  <* pN
        e = ((\i n -> if n == 1 then 1 else div' i n + div' 1 n) <$> pI) <* pN
        div' num den = fromIntegral (num `mod` den) / fromIntegral den
```

Then I tried patterning the sample width to calculate the end parameter pattern:

```haskell
slice'' :: Pattern Int -> Pattern Int -> ControlPattern -> ControlPattern
slice'' pN pI p = P.begin b |> P.end e |> p
  where
    b = div' <$> pI <* pN
    e = b + pWidth
    pWidth = (\x -> 1.0 / fromIntegral x) <$> pN
    div' :: Int -> Int -> Double
    div' num den = fromIntegral (num `mod` den) / fromIntegral den
```

It looks like this issue is also true for `splice`:

> λ> splice 1 0 $ s "bev"
> (0>1)|begin: 0.0f, end: 0.0f, s: "bev", speed: 1.0f, unit: "c"

I can put a PR together if this will work with tidal 2.0

yaxu:
Hi @grddavies, a PR would be very welcome. Please target the `dev` branch, thanks!

PR #1086, opened 2024-06-04
fix(slice): support `slice 1 _`

grddavies:
fixes: #979

Slice was failing on any slice `pN` of 1. This was because the `begin` of the next slice was being used to calculate the end of the first. With `slice 1 _` all slices begin at `0.0`, so end is incorrectly set to `0.0`.

This implementation calculates the `end` control pattern by generating a sample-width pattern, and adding this to the `begin` pattern.

Two tests are added:

* `slice` "can slice samples" - Regression test that passes on `dev` and this branch
* `slice` "can slice by 1" - Test for accurate calculation of sample end time when slicing into 1


yaxu:
Thanks!

Issue #1089, opened 2024-06-16
setcps = once . cps is broken


polymorphicengine:
for now, please use `streamSetCPS` instead.

this bug is likely due to something in `onSingleTick` / `Clock.getZeroedLinkOperations`.

PR #1090, opened 2024-06-16
simplify the clock logic by removing LinkOperations

polymorphicengine:
it was pretty hard to track down what was causing #1089, so i found it easier just to remove the LinkOperations datatype and replace it with a bunch of functions that act on a given session state / a clock reference / the clock config

i think this is a good next step in simplifying the stream/clock logic. 

tested everything and seems to be fine, but please have a look @Zalastax, @mindofmatthew, @yaxu :)

### DIFF
--- a/src/Sound/Tidal/Stream/Process.hs
+++ b/src/Sound/Tidal/Stream/Process.hs
@@ -34,7 +34,6 @@ import           Control.Monad             (forM_, when)
 import           Data.Coerce               (coerce)
 import qualified Data.Map.Strict           as Map
 import           Data.Maybe                (catMaybes, fromJust, fromMaybe)
-import           Foreign.C.Types
 import           System.IO                 (hPutStrLn, stderr)
 
 import qualified Sound.Osc.Fd              as O
@@ -47,7 +46,6 @@ import qualified Sound.Tidal.Link          as Link
 import           Sound.Tidal.Params        (pS)
 import           Sound.Tidal.Pattern
 import           Sound.Tidal.Show          ()
-import           Sound.Tidal.Stream.Config
 import           Sound.Tidal.Utils         ((!!!))
 
 import           Sound.Tidal.Stream.Target
@@ -57,7 +55,7 @@ data ProcessedEvent =
   ProcessedEvent {
     peHasOnset         :: Bool,
     peEvent            :: Event ValueMap,
-    peCps              :: Link.BPM,
+    peCps              :: Double,
     peDelta            :: Link.Micros,
     peCycle            :: Time,
     peOnWholeOrPart    :: Link.Micros,
@@ -88,9 +86,11 @@ doTick :: MVar ValueMap                           -- pattern state
        -> Maybe O.Udp                             -- network socket
        -> (Time,Time)                             -- current arc
        -> Double                                  -- nudge
-       -> Clock.LinkOperations                    -- ableton link operations
+       -> Clock.ClockConfig                       -- config of the clock
+       -> Clock.ClockRef                          -- reference to the clock
+       -> Link.SessionState
        -> IO ()
-doTick stateMV busMV playMV globalFMV cxs listen (st,end) nudge ops =
+doTick stateMV busMV playMV globalFMV cxs listen (st,end) nudge cconf cref ss =
   E.handle (\ (e :: E.SomeException) -> do
     hPutStrLn stderr $ "Failed to Stream.doTick: " ++ show e
     hPutStrLn stderr $ "Return to previous pattern."
@@ -99,10 +99,10 @@ doTick stateMV busMV playMV globalFMV cxs listen (st,end) nudge ops =
       pMap <- readMVar playMV
       busses <- readMVar busMV
       sGlobalF <- readMVar globalFMV
-      bpm <- (Clock.getTempo ops)
+      bpm <- Clock.getTempo ss
       let
         patstack = sGlobalF $ playStack pMap
-        cps = ((Clock.beatToCycles ops) bpm) / 60
+        cps = ((Clock.beatToCycles cconf) $ fromRational bpm) / 60
         sMap' = Map.insert "_cps" (VF $ coerce cps) sMap
         extraLatency = nudge
         -- First the state is used to query the pattern
@@ -112,7 +112,7 @@ doTick stateMV busMV playMV globalFMV cxs listen (st,end) nudge ops =
                                                 )
          -- Then it's passed through the events
         (sMap'', es') = resolveState sMap' es
-      tes <- processCps ops es'
+      tes <- processCps cconf cref ss es'
       -- For each OSC target
       forM_ cxs $ \cx@(Cx target _ oscs _ _) -> do
               -- Latency is configurable per target.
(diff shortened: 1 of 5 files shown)


Zalastax:
This is a tricky area and it's good that you try to simplify!

By using setCPS, the change will not be effective until next cycle. This could cause a difference in the performance. What do you think about this potential difference?

polymorphicengine:
thanks for looking into it! 
are you sure it wont be effective until the next cycle?
when i try it, it seems instant. i'm doing something like

```
ghci> import Sound.Tidal.Context 
ghci> tidal <- startTidal superdirtTarget defaultConfig 
ghci> :set -XOverloadedStrings 
ghci> streamReplace tidal 1 $ s "ho sn bd sn"
ghci> streamOnce tidal $ cps 2
```

src/Sound/Tidal/Stream/Process.hs
```diff
@@ -124,27 +124,29 @@ doTick stateMV busMV playMV globalFMV cxs listen (st,end) nudge ops =
...
-          offBeat = (Clock.cyclesToBeat ops) (realToFrac offCycle)
-      on <- (Clock.timeAtBeat ops) onBeat
-      onPart <- (Clock.timeAtBeat ops) partStartBeat
+          offBeat = (Clock.cyclesToBeat cconf) (realToFrac offCycle)
+      on <- Clock.timeAtBeat cconf ss onBeat
+      onPart <- Clock.timeAtBeat cconf ss partStartBeat
       when (eventHasOnset e) (do
         let cps' = Map.lookup "cps" (value e) >>= getF
-        maybe (return ()) (\newCps -> (Clock.setTempo ops) ((Clock.cyclesToBeat ops) (newCps * 60)) on) $ coerce cps'
+        maybe (return ()) (\newCps -> Clock.setCPS cconf cref newCps) (fmap toRational cps')
```
Zalastax, line 143: I expect this to behave differently? Could cause subtle differences when events change cps and the events coming afterwards won't use the new cps until next tick.

Zalastax:
Yes, I am quite confident. This is because setCPS just adds a SetTempo entry to the actions list. The actions are processed between the ticks. So with this change, events that use cps won't immediately affect the timing. I think this could be problematic for some people, while most won't notice.

src/Sound/Tidal/Stream/Process.hs
polymorphicengine: yes it might be a difference of a tick, the reason i did here like this is to avoid carrying around an additional sessionstate that would keep track of time changes (this was previously possible, since the LinkOperations would carry this specific version of setTempo)
i'm not so sure if the complication is worth it here

polymorphicengine:
hmm now that i think of it, would a version of setTempo work that creates a new session state, calls Link.setTempo and commits the session state?

Zalastax:
> hmm now that i think of it, would a version of setTempo work that creates a new session state, calls Link.setTempo and commits the session state?

It could possibly work if it also calls setTempo on the original session state that's used for the tick. This is needed since commits on the new session state don't affect the original. For the single tick case I see no issues because the original session state is not committed but I don't know how the normal tick case is affected when we set tempo in two session states and commit both.

src/Sound/Tidal/Stream/Process.hs
Zalastax: I'm not really active in this project at the moment so someone else need to make the call of whether the complexity is worth it. But now you see the motivation for the current design.

tidal-link/src/hs/Sound/Tidal/Clock.hs
```diff
@@ -249,8 +222,35 @@ arcStart = fst
...
+
+-- onSingleTick assumes it runs at beat 0.
+-- The best way to achieve that is to use forceBeatAtTime.
+-- But using forceBeatAtTime means we can not commit its session state.
+getZeroedSessionState :: ClockConfig -> ClockRef -> IO Link.SessionState
+getZeroedSessionState config (ClockRef _ abletonLink) = do
+                            ss <- Link.createAndCaptureAppSessionState abletonLink
+                            nowLink <- liftIO $ Link.clock abletonLink
+                            Link.forceBeatAtTime ss 0 (nowLink + processAhead) (cQuantum config)
+                            Link.destroySessionState ss
```
Zalastax, line 239: It seems unsafe to use ss after it is destroyed. If we want to ensure ss is freed, then I think we need to move flow control into this function. Here we could accept a function (Link.SessionState -> IO a) as argument and return IO a. We would create ss, run the function, then destroy ss, then return.

src/Sound/Tidal/Stream/Process.hs
```diff
@@ -140,7 +140,7 @@ processCps cconf cref ss = mapM processEvent
       onPart <- Clock.timeAtBeat cconf ss partStartBeat
       when (eventHasOnset e) (do
         let cps' = Map.lookup "cps" (value e) >>= getF
-        maybe (return ()) (\newCps -> Clock.setCPS cconf cref newCps) (fmap toRational cps')
+        maybe (return ()) (\newCps -> Clock.setTempoCPS newCps on cconf ss) (fmap toRational cps')
```
Zalastax, line 143: Doesn't this mean we will lose the tempo changes that happen in the single tick case since we don't (and can't) commit the zeroed session state?

tidal-link/src/hs/Sound/Tidal/Clock.hs
```diff
@@ -288,42 +295,10 @@ getCycleTime :: ClockConfig -> ClockRef -> IO Time
...
-getZeroedLinkOperations config (ClockRef _ abletonLink) = do
-              sessionState <- Link.createAndCaptureAppSessionState abletonLink
-              zeroedSessionState <- Link.createAndCaptureAppSessionState abletonLink
-
-              nowOsc <- O.time
-              nowLink <- Link.clock abletonLink
-
-              Link.forceBeatAtTime zeroedSessionState 0 (nowLink + processAhead) (cQuantum config)
-
-              Link.commitAndDestroyAppSessionState abletonLink sessionState
```
Zalastax, line 310: Could this be the source of the bug? We are running commitAndDestroy immediately - before the IO where the session states are used. Compare to clockProcess where I understand we first process the arc using liftIO, then we run commitAndDestroy.

tidal-link/src/hs/Sound/Tidal/Clock.hs
Zalastax: Alternatively, could we move the destroySessionState into a callback function which we include in the return? New return type becomes `IO (Link.SessionState, IO ())`.

tidal-link/src/hs/Sound/Tidal/Clock.hs
Zalastax: The suggestion in https://github.com/tidalcycles/Tidal/pull/1090/files/81c19870a81294e4b5def9cdc24365fa0b23e726#r1645196629 could be applied here. The callback would invoke both destroy functions.

src/Sound/Tidal/Stream/Process.hs
polymorphicengine: i thought so too, but `streamOnce tidal $ cps 1` is still working.. not completely sure why?

tidal-link/src/hs/Sound/Tidal/Clock.hs
polymorphicengine: hmm yes that makes sense, the callback is probably a good idea! thanks!

polymorphicengine:
this should be good to go now. 

i introduced an additional session state to keep track of tempo changes:
in case of the `doTick` call inside of the running clock, it will just be twice the same session state, but in case of `doSingleTick` it will be a zeroed session state and the new additional one.

also, as you suggested, they are deleted after the call to `doTick` now.

do let me know what you think @Zalastax !

Zalastax:
This looks good to me!

I think there could be some more encapsulation so that Clock is in charge of all the calls to Link, but I don't find that so important. I would be OK with merging this.

polymorphicengine:
> I think there could be some more encapsulation so that Clock is in charge of all the calls to Link

yes i agree - added one more commit that moves the code to a function `clockOnce` in the Clock module that you simply pass a `TickAction`.

also bumped tidal-links version to `1.0.4`

 i think this is good to go now!

mindofmatthew:
This all looks good to me, although I admit I'm not very knowledgeable about the Link integration.

A direction I'd like to see with encapsulation is the ability to swap in a deterministic mock clock for writing more Stream tests. That can definitely wait until a future PR of course.

yaxu:
Will try it out in the wild, thanks!

Issue #761, opened 2021-01-18
Contiguous anything?


jwaldmann:
Reading #760 got me thinking:

Could we extract this "contiguous version of" (a.k.a. `gapless`) as a function for the Haskell API? 

So that `euclidGapless x y = gapless $ euclid x y (pure True)` or similar.

It seems that this hypothetical `gapless`  can only be applied to sequences, not arbitrary patterns - an instance of https://club.tidalcycles.org/t/accumulation-over-patterns-why-we-cannot-have-it/2706 ?

This indicates to me that `Pattern` is too abstract for some applications. In Haskell land, we can then use the list type `[a]` but in mininotation, everything's a pattern, except sometimes it's not ( https://club.tidalcycles.org/t/polymetrics-api-fication/2705 )

PR #1093, opened 2024-07-19
encloseEvent & query windowing functions

geikha:
**This PR is a work in progress, it hasn't been tested with the new tactus functionallity**

Related: https://github.com/tidalcycles/strudel/pull/1142

With this PR I try to introduce a "new" way of time altering functions to Tidal. A method which I'll be calling windowing. This method is actually done in some ways by other functions in some way, but this PR simplifies the process and introduces new functions. The idea is that for every query with a given state arc, Tidal queries an expanded window (for example, the whole cycleArc surrounding the query, or by some lookahead), processes the events found and then windows it back down to the original arc via the use of the new `encloseEvent` function. This function takes an Arc and an Event, and returns Just the Event as if it was queried from the given Arc, or Nothing (if it shouldn't be there in the first place). This makes it easy to implement functions which alter times of events between a single cycleArc.

UI functions added:
* `quant k`: quantizes the times of events to steps of (1/k) cycles. Inputting 0 will bypass the effect.
* `fill m`: fills gaps between events, essentially a proper legato (name being debated @ https://github.com/tidalcycles/strudel/pull/1142)
* `alterT f`: takes a Time modifying function to alter the cyclePos of the events
* `alterFf`: sames as above but uses a Double function

The `alter` functions have some buggy edge cases I'm trying to explore (events at the exact end of the cycle, negative results, etc).

`fill` sorts events just in case, but i have no idea if this actually necessary on tidal (it is on strudel, i was replicating jader's work with the events on this one)

I've noted there's an old TODO `fill` function on the Pattern package, I think it should be easier to implement with the new functions... I'll take a look.

---

TODO:

- [ ] Decide on a name for the proposed `fill`
- [ ] How should `encloseEvent` actually handle analog events?
- [ ] Test while using tactus functions

### DIFF
--- a/src/Sound/Tidal/Pattern.hs
+++ b/src/Sound/Tidal/Pattern.hs
@@ -502,6 +502,16 @@ withEvents f p = p {query = f . query p, pureValue = Nothing}
 withPart :: (Arc -> Arc) -> Pattern a -> Pattern a
 withPart f = withEvent (\(Event c w p v) -> Event c w (f p) v)
 
+-- | @withEventsOnArc ef af p@ returns a new @Pattern@ with ef applied to the events list queried from the query arc modified by af, then enclosed into the original arc
+-- function @f@
+withEventsOnArc :: ([Event a] -> [Event a]) -> (Arc -> Arc) -> Pattern a -> Pattern a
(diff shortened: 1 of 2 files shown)


geikha:
Would fix https://github.com/tidalcycles/Tidal/issues/761

PR #1098, opened 2024-11-05
flake.nix: update nixpkgs and remove temporary hosc input

ahihi:
hosc 0.20 is now in nixpkgs-unstable, so the extra input no longer required.


yaxu:
thanks!

PR #1100, opened 2024-11-05
doTick: use exception-safe modifyMVar_ instead of takeMVar/putMVar

### LINKED ISSUE #1099: mini-notation syntax error causes stream to freeze
sometimes, a syntax error in the mini-notation causes the stream thread to freeze. an example of a pattern that causes this is:

```haskell
d1 $ stack
  [ (1/16) ~> s "ride*4"
  , n "<0 1 2§>*16" # s "808" -- note the invalid '§'
  ]
```

after this, there is no sound even if i evaluate a valid block like `d1 $ s "bd*4"`.

ahihi:
fixes #1099, maybe.

disclaimer: i dont fully understand the logic of manipulating `stateMV` in `doTick` (and the other locations that touch this MVar)... at the beginning the value is taken out, and if everything runs successfully, a new value is put in. the exception handler doesnt touch `stateMV`, so it seems like any exception should leave it empty. yet i dont see the hang with all mini-notation syntax errors - for example, Tidal recovers just fine from `d1 $ n "<0 1 2§>*16" # s "808"`. not sure how.

but i am making a guess that the intent here is not to leave `stateMV` empty on error. to this end, i replaced the `takeMVar`/`putMVar` sequence with a `modifyMVar_`. from my testing so far, this seems to resolve the issue, but im not sure if there might be some undesired side effects.

(i also moved the exception handler into its own function, as i found the previous parenthesized multi-line blocks difficult to read, but let me know if that style is preferred)

### DIFF
--- a/src/Sound/Tidal/Stream/Process.hs
+++ b/src/Sound/Tidal/Stream/Process.hs
@@ -91,11 +91,8 @@ doTick :: MVar ValueMap                           -- pattern state
        -> (Link.SessionState, Link.SessionState)  -- second session state is for keeping track of tempo changes
        -> IO ()
 doTick stateMV busMV playMV globalFMV cxs listen (st,end) nudge cconf cref (ss, temposs) =
-  E.handle (\ (e :: E.SomeException) -> do
-    hPutStrLn stderr $ "Failed to Stream.doTick: " ++ show e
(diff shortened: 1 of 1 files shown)


yaxu:
Thanks, lets merge for now and try it out!

Issue #1035, opened 2023-09-12
OSC Playback: Add setcps control message 


jarmitage:
References: 
- https://tidalcycles.org/docs/configuration/MIDIOSC/osc/#playback-controllers
- https://github.com/tidalcycles/Tidal/blob/fcc4c5d53a72dcf2b8f4c00cc1c1b3c75eef172d/src/Sound/Tidal/Stream.hs#L699

Would be great to add `setcps` to this list of control input messages.

sss-create:
feature added with #1102 and will be in tidal 1.10! 

closing now :)

PR #1102, opened 2025-01-11
#1035 added setcps and other controls to OSC Listener

sss-create:
As the title says.. :)
Tested by sending OSC messages to the REPL.

@yaxu, I hope I am not being intrusive by going through some issues that may be unimportant or abandoned for good.


yaxu:
Thanks looks good!  It seems the tests are broken at the moment but will merge for now..

Issue #1103, opened 2025-01-11
hosc-0.21 breaking changes


sss-create:
hosc-0.21 apparently breaks tidal-listener since Functions and Types are now differently created in hosc. E.g. ```data OscProtocol```. Maybe Tidal should revert and expect hosc 0.20 for now and either stay there or incorporate the hosc changes into tidal-listener and Tidal itself, which I would volunteer for to do so -- if desired.

yaxu:
Yes agreed we should stick to hosc-0.20. 

There's also a tighter version dependency in `hosc-0.21` requiring a more recent ghc than the one in current debian, that I need to get to the bottom of.

It would be great to upgrade to hosc-0.21 at some point though, as we're currently preventing its release on stackage https://github.com/commercialhaskell/stackage/issues/7627

Rohan actually uploaded hosc-0.21 on my request, to get rid of an obsolete dependency that caused tidal to be removed from debian: 
  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1083252

yaxu:
Oops only just read your post properly.. Yep just putting a lower bounds on bytestring on hosc seems to fix it! I'm pretty sure I tried setting the lower bounds in tidal and this didn't work, so I just assumed bytestring was part of core and couldn't be upgraded.. I'll send rohan another email!

yaxu:
> edit: So implementing hosc-0.21/0.22 in Tidal could free the way for hosc into stackage? Is this, plus maybe the bipolar LFO feature you mentioned in #1101, enough for a new Tidal release?

Yes there's quite a lot of other changes waiting for release.

## What's Changed
* replace Tempo module by an ableton-link synched clock that comes with tidal-link by @polymorphicengine in https://github.com/tidalcycles/Tidal/pull/1059
* Add capital-M aliases for major chords by @yaxu in https://github.com/tidalcycles/Tidal/pull/1077
* Cherry pick new Boot changes and adjust to incorporate recent changes by @matthewkaney in https://github.com/tidalcycles/Tidal/pull/1081
* bump upper bounds of network dependency to 3.3 by @yaxu in https://github.com/tidalcycles/Tidal/pull/1080
* more efficient sew, plus some auto-reformatting by @yaxu in https://github.com/tidalcycles/Tidal/pull/1078
* Step-oriented functionality by @yaxu in https://github.com/tidalcycles/Tidal/pull/1072
[...]

PR #1105, opened 2025-01-15
hosc-0.21 support

sss-create:
I probably won't be able to keep this tempo up. 😅

However, this compiles and was tested by sending patterns from the REPL to SuperCollider. Tidal is still controllable via OSC (tested with Pure Data) too.

edit: would solve #1103

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,29 @@
 {
   "nodes": {
+    "hosc": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1734661491,
+        "narHash": "sha256-ilu/WjulVrEtP+tBwTf2kughoziWhEm27MigkOm9F6c=",
+        "owner": "rd--",
+        "repo": "hosc",
+        "rev": "43bb2d07ff8d65cf9e51d1f5f96d0e6ffd6fe8fa",
+        "type": "github"
+      },
+      "original": {
+        "owner": "rd--",
+        "repo": "hosc",
+        "rev": "43bb2d07ff8d65cf9e51d1f5f96d0e6ffd6fe8fa",
+        "type": "github"
+      }
+    },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1730531603,
-        "narHash": "sha256-Dqg6si5CqIzm87sp57j5nTaeBbWhHFaVyG7V6L8k3lY=",
+        "lastModified": 1737062831,
+        "narHash": "sha256-Tbk1MZbtV2s5aG+iM99U8FqwxU/YNArMcWAv6clcsBc=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "7ffd9ae656aec493492b44d0ddfb28e79a1ea25d",
+        "rev": "5df43628fdf08d642be8ba5b3625a6c70731c19c",
         "type": "github"
       },
       "original": {
@@ -18,6 +35,7 @@
     },
     "root": {
       "inputs": {
+        "hosc": "hosc",
         "nixpkgs": "nixpkgs",
         "utils": "utils"
       }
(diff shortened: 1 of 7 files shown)


sss-create:
Ah! CI shows, there is still a problem with the bytestring version dependency:
the unix package expects ```0.10.10.1``` exclusively!
https://github.com/tidalcycles/Tidal/actions/runs/12786157753/job/35642668849?pr=1105
``` rejecting: hosc-0.21 (conflict: unix =>
bytestring==0.10.10.1/installed-0.10.10.1, hosc => bytestring>=0.11.0.0)```

yaxu:
Ah, that explains why it still isn't compiling for me. I was hoping increasing lower bounds for `binary` would get me out of dependency hell but it seems not:
  https://hackage.haskell.org/package/hosc-0.21/revisions/

I think Rohan is happy to change the dependencies for hosc on hackage further (this is possible without doing a new hosc release). But I'm not sure if this helps?

Perhaps we just have to say that ghc 9.x has been around for several years and we drop support for 8.x Normally I use a recent ghc via ghcup, I just happened to be using the system version on my laptop for some reason.

Debian stable is on ghc 9.x, my laptop install (some linux mint) must be based on oldstable.
  https://packages.debian.org/search?keywords=ghc

@cleary do you have a view on this wrt the ansible-based install?

sss-create:
@yaxu While dropping support for 8.x seems reasonable, I'm not sure if 9.0.2 has a similar error regarding at least the unix package: 
https://github.com/tidalcycles/Tidal/actions/runs/12786157753/job/35642667844?pr=1105

`unix` expects `0.10.12.1` there, where `fromStrict` is still not exported from `Data.ByteString`

yaxu:
Ok have asked Rohan about applying the patch for backward compatibility

see you

yaxu:
hosc 0.21.1 released! seems a bit happier now?

sss-create:
I‘d guess the Nix builds fail because there is no hosc-0.21.1 Nix package?
https://search.nixos.org/packages?channel=24.11&from=0&size=50&sort=relevance&type=packages&query=hosc

I just wonder why the working Nix build is able to use hosc 0.20, although we specify >0.20.

yaxu:
I've not got my head around nix at all I'm afraid..

sss-create:
Me neither. But by asking some questions on the haskell irc I found out, that there is a script which should update haskell modules in Nix every few hours or at least auto generate PRs that will be merged every few days. So probably waiting one or two days will solve this for us. :)

ahihi:
i am using the nix flake, and am happy to test and update it again once hosc 0.21 reaches nixpkgs

yaxu:
Great thanks! Time to merge?

sss-create:
Depends what is more important, I guess. To get a new Tidal release done to free hosc‘s way into stackage or don‘t break the Nix package/install. As I understand only one can be chosen currently? In favor of merging an preparing a release could be that Nix also looks at Stackage packages, when Hackage packages are not found — but I am really not sure about the whole Nix process.

ahihi:
while we wait for nixpkgs, there is also the option of bringing back the hosc override in flake.nix that was removed in #1098 (updated to 0.21, of course)

yaxu:
I don't know about nix, but I understand everything on stackage comes from hackage. It's like a curated set of hackage package version numbers that work together. So we upload to stackage, then do a PR to remove the upper bounds on hosc in stackage. No rush anyway, tidal is the only thing using hosc on stackage.

yaxu:
Maybe @mitchmindtree has advice as maintainer of https://github.com/mitchmindtree/tidalcycles.nix

mitchmindtree:
Ahh I think @ahihi's right, the easiest approach might be to re-add the `hosc` source to the flake - I've just opened a small PR against this branch here: https://github.com/sss-create/Tidal/pull/1

yaxu:
Thanks all !

PR #1106, opened 2025-01-17
Fix off-by-one error in striateBy

yaxu:
Also make it work within any existing begin/end controls, like `striate` does


yaxu:
Ref discussion here: https://club.tidalcycles.org/t/chopby/5210/4?u=yaxu

Issue #1041, opened 2023-09-30
Executable name conflict of "linktest"


felixonmars:
Executable with the same name exists in the mesa demos. This creates conflicts when the user attempts to install both on the same device. It would be a good idea if we consider to rename the executable.

PR #1107, opened 2025-01-18
renamed linktest executable

sss-create:
This would resolve issue #1041.... I just don't know if this would break anything. I assume it won't but still.. -- hence the PR and no direct push. 😶‍🌫️

### DIFF
--- a/tidal-link/tidal-link.cabal
+++ b/tidal-link/tidal-link.cabal
@@ -69,7 +69,7 @@ source-repository head
   type:     git
   location: https://github.com/tidalcycles/Tidal
 
-executable linktest
+executable tidal-linktest
   ghc-options: -Wall
   hs-source-dirs:
                  src/hs


yaxu:
I don't see why this would be a problem, but @Zalastax will be more familiar with it

yaxu:
BTW I think best we do PRs for everything anyway, even when you merge yourself, not least because it makes release notes easier.

sss-create:
Ok, got it :)

While we're at it: Do we need to host the Ableton Link code ourselves here or could we reference the ableton repo as a submodule (with a specific commit)? Maybe @Zalastax can answer this too.

Zalastax:
The renaming should be fine.

I used submodules originally but we changed to use file copy: https://github.com/tidalcycles/Tidal/pull/891#issuecomment-1035635790

I don't see much benefit to change back to submodules but I think it might be time to update the files.

sss-create:
The benefits I see are:

* get rid of 500k lines of C++ code we don’t edit or need to work with
* make visible that tidal-link depends on the link repo
* don’t have to manually update the files
* make the overall Tidal repo code (I assume) get rid of the >90% C++ code indicator, which made me question if Tidal was really written in Haskell 

However submodules are or can be fiddly, that‘s true.

yaxu:
No strong feelings either way, but appreciate those benefits over fiddlyness.

I think we should keep up-to-date with upstream as much as possible, in the past we lost compatibility with new mac os versions due to being behind. I imagine that's easier to track with submodules. I think fine to go ahead if you want to try it @sss-create.

sss-create:
We can merge and close the issue too. I‘ll open a new PR later on. :)

PR #1109, opened 2025-01-20
`beat` function for "step sequencer" style rhythm notation. (Backported from Strudel)

geikha:
See https://github.com/tidalcycles/strudel/pull/1237

Luckily the fractional implementation done by @yaxu doesn't require any special handling of polyphony, it just works!

```hs
beat :: Pattern Time -> Pattern Time -> Pattern a -> Pattern a
beat = patternify2 $ __beat innerJoin

__beat :: (Pattern (Pattern a) -> Pattern a) -> Time -> Time -> Pattern a -> Pattern a
__beat join t d p = join $ (compress (s,e) . pure) <$> p
                      where s = t' / d
                            e  = (t'+1) / d
                            t' = t `mod'` d
```

### DIFF
--- a/src/Sound/Tidal/UI.hs
+++ b/src/Sound/Tidal/UI.hs
@@ -1461,6 +1461,23 @@ _markovPat n xi tp = setTactus (toRational n) $ splitQueries $ pattern (\(State 
   queryArc (listToPat $ runMarkov n tp xi (sam s)) a)
 
 {-|
+@beat@ structures a pattern by picking subdivisions of a cycle.
+Takes in a pattern that tells it which parts to play (polyphony is recommeded here),
+and the number of parts by which to subdivide the cycle (also pattern-able).
+For example:
+> d1 $ beat "[3,4.2,9,11,14]" 16 $ s "sd"
+-}
+beat :: Pattern Time -> Pattern Time -> Pattern a -> Pattern a
+beat = patternify2 $ __beat innerJoin
+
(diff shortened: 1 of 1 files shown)


sss-create:
Wow, this looks interesting! Can you tell me what it does? I did not quite get the hang of it in the strudel thread + I couldn’t listen to the audio

yaxu:
`struct (beat "[0,5,12]" "16") $ s "bd"` would trigger on the first, sixth and 13th beat of a 16 beat cycle. It's a bit weird that the beats are 'stacked' like this but makes some sense..

geikha:
if anyone wants to try this on the current version of tidal without installing anything you can just run:

```hs
__beat :: (Pattern (Pattern a) -> Pattern a) -> Time -> Time -> Pattern a -> Pattern a
__beat join t d p = join $ (compress (s,e) . pure) <$> p
                    where s = t' / d
                          e  = (t'+1) / d
                          t' = t `Data.Fixed.mod'` d
beat :: Pattern Time -> Pattern Time -> Pattern a -> Pattern a
beat = tParam2 $ __beat innerJoin
```

yaxu:
> @yaxu i think you meant `beat "[0,5,12]" "16" $ s "bd"`, as the function restructures the pattern already
> 
> also, i like that we can easily `beat "[0,5,<12 11>]" 16` and stuff like that

Ah hmm a binary version might be more flexible, but lets park that thought for now and merge :)

cleary:
I think of this like a flexible Euclidean notation - rather than evenly spreading over the increments you just define where you want them to fall.

Its easier than counting out the steps in binary notation too

PR #1110, opened 2025-01-20
remove extra paren in tidal.el

jezcope:
A typo in c278266a96a8d2c00ca670b1b0b3eb777d3f45ae added an extra `)` on line 349 of `tidal.el` which meant that the file wouldn't parse and broke `tidal-mode`. This fixes it.


jezcope:
PS. thanks for this cool software, and including an emacs mode for it!

sss-create:
Oh shoot! Thank you!

PR #1112, opened 2025-01-22
Reformat

yaxu:
My editor reformatted these files on the patterned-tactus branch without me realising. I could try to untangle my changes and undo the reformat, but thought I'd reformat them separately on this PR, and then merge back into patterned-tactus, to hopefully end up with a clean PR for that.

### DIFF
--- a/src/Sound/Tidal/Bjorklund.hs
+++ b/src/Sound/Tidal/Bjorklund.hs
@@ -22,29 +22,29 @@ module Sound.Tidal.Bjorklund (bjorklund) where
 -- the library but removed for now due to dependency problems.. We
 -- could however likely benefit from other parts of the library..
 
-type STEP a = ((Int,Int),([[a]],[[a]]))
+type STEP a = ((Int, Int), ([[a]], [[a]]))
 
 left :: STEP a -> STEP a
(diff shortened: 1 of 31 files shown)


yaxu:
Looking into it, this is formatted with ormolu, which has clear goals stated here:
  https://github.com/tweag/ormolu

It seems to be purposefully non-configurable to create some kind of standard way of formatting.

Previously I've used stylish:
  https://github.com/haskell/stylish-haskell

Without doing a close comparison, is seems to be more "light touch".

There is also fourmolu, which is a fork of ormolu that is more configurable. It's called fourmolu from its (default?) four-character indent.
  https://github.com/fourmolu/fourmolu

These are all actively maintained, and fourmolu merges upstream changes from ormolu.

I have no strong feelings on code formatting rules, so would be happy to go with a opinionated option like ormolu as I don't have strong opinions of my own. :) What do others think?

I think we should try to merge as many PRs as possible before applying the formatting to all files.

ndr-brt (commented):
just a comment out of the blue (I'm still lurking all the PRs :) ) 
I like this formatting style a lot :rocket:

Issue #1091, opened 2024-07-01
Parser errors for new rest symbol


nkleemann:
As a german keyboard layout user, thank you very much for: #1087 

After experimenting with this I found patterns that trips up the parser: 

* "[--" 
* "<---"
* "- 2" (as used in `$ slow "- 2"`, `$ slow "~ 2"` is working fine)

example:

```
d1
    $ n "[-- 0 5]"
    # s "cpu"
```

Failed to Stream.streamReplace: Syntax error in sequence:
  "[-- 0 5]"
           ^  
unexpected end of input
expecting sequence, ",", "|" or "]"
Return to previous pattern.

yaxu:
Ah thanks for this, the patch was surprisingly simple, I guess too simple! I'll add some breaking tests to mark this out as a release blocking bug

sss-create:
I wanted to look at the parser anyways :)

yaxu:
> I wanted to look at the parser anyways :)

Thanks @sss-create , it looks like I failed to add the test but it is still broken. I added it to a milestone for the next release 1.10 for now.

PR #1113, opened 2025-01-23
fixed: Parser errors for new rest symbol #1091

sss-create:
as described in #1091 

the 'edge cases' resolved were:
1. the '-' could be a negative number or part of a rational number.
2. the angles and brackets tokenization from native Parsec behave weird.

puh! this was a tough one for me to find because of point 2.

Tests work, so if @yaxu gives his ok, this can be merged.

### DIFF
--- a/src/Sound/Tidal/ParseBP.hs
+++ b/src/Sound/Tidal/ParseBP.hs
@@ -41,7 +41,7 @@ import           Data.Typeable                          (Typeable)
 import           GHC.Exts                               (IsString (..))
 import           Sound.Tidal.Chords
 import           Sound.Tidal.Core
-import           Sound.Tidal.Pattern
+import           Sound.Tidal.Pattern hiding             ((*>), (<*))
 import           Sound.Tidal.UI
 import           Sound.Tidal.Utils                      (fromRight)
 import           Text.Parsec.Error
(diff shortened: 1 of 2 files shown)


yaxu:
I'm happy for it to be merged, but perhaps worth adding a test or two to `test/Sound/Tidal/ParseTest.hs` if you have time?

sss-create:
I did. Interesting: 
comparing this with itself leads to a failing test. Do I correctly use 3%4 here?
``
("[~~ 2 <~~ 2@7 3> 3%4 9|8 ~~ [~~ <2 9q> ~]] 2!4" :: Pattern String)
``

Comparing and using rationals like in the added test passes it.

yaxu:
Hmm, well values `9q` and `3%4` look like they are expecting the parser for `Pattern Rational` which is parsed differently to `Pattern String`.

So I would try changing `Pattern String` for `Pattern Rational` and see if the tests pass then.

But still, although parsing this as `Pattern String` would lead to difficult to anticipate results, those results should at least be deterministic, so don't know what's going on with that test..

sss-create:
Ah yes! On both branches — with and without my changes — the test using `Pattern String` does fail, while the test parsed as `Pattern Rational` passes.

yaxu:
Hmm, if I check out your dev branch, the tests pass, and this comparison succeeds:
```hs
(queryArc ("[-- 2 <-- 2@7 3> 1*4%2 3? 4 9|8 -- [-- <2 9q> -]] 2!4" :: Pattern String) (Arc 0 127)) == (queryArc ("[~~ 2 <~~ 2@7 3> 1*4%2 3? 4 9|8 ~~ [~~ <2 9q> ~]] 2!4" :: Pattern String) (Arc 0 127))
```

yaxu:
If I run that on the main tidal dev branch, I get a syntax error from parsec..

yaxu:
I think it's probably working fine!

sss-create:
I added a test for ```Pattern Rational```. Will merge this now :)

Issue #1092, opened 2024-07-08
tidal-listener first.hs example not working


ndr-brt:
When I try to go through the [first.hs](https://github.com/tidalcycles/Tidal/blob/dev/tidal-listener/examples/first.hs) example in the `tidal-listener/examples` folder I get some errors.

Some of them are related to a wrong case in the `Sound.Osc.Fd` library calls, but, when I try to evaluate the first instruction (`return 10`), as response message I get:
```
Just (Message {messageAddress = "/eval/error", messageDatum = [AsciiString {d_ascii_string = "Compile error: Could not load module \CANSystem.Directory\EM.\nIt is a member of the hidden package \CANdirectory-1.3.8.1\EM.\nUse -v to see a list of the files searched for."}]})
```

sss-create:
I can reproduce. Running 3 terminals:

```
$ cabal run
Starting Tidal Listener with installed GHC
Listening for OSC commands on port 6011
Sending replies to port 6012
[TidalCycles version 1.10.0]
Installed in /home/..../Tidal/./.
Listening for external controls on 127.0.0.1:6010
Starting tidal interpreter..
Waiting for SuperDirt (v.1.7.2 or higher)..
```

and later: ``` tidal-listener: thread blocked indefinitely in an MVar oper```

one ghci to send:
```
ghci> openUdp "127.0.0.1" 6011 >>= \r -> sendMessage r $ Message "/eval" [string "return 10"]
```

and a last one for receiving/reading the msg: 

```
ghci> recvMessage udp
Just (Message {messageAddress = "/eval/error", messageDatum = [AsciiString {d_ascii_string = "Compile
error: Could not find module \CANSound.Tidal.Transition\EM.\nUse -v to see a list of the files searched for."}]})
```

I will investigate further in the next days.

PR #1114, opened 2025-01-23
CI on tidal-listener execute on changes only

sss-create:
As I understand tidal-listener is not used in Tidal or tidal-link itself and currently a bit broken as mentioned in #1092.

Therefore it does not make sense to me to run the full CI (which works towards a packaged release, I think) on every change in Tidal or tidal-link. If I'm wrong, feel free to just close this :)

### DIFF
--- a/.github/workflows/listener-build-linux.yml
+++ b/.github/workflows/listener-build-linux.yml
@@ -1,11 +1,11 @@
 name: build-listener-linux
 on:
   push:
-    paths-ignore:
-      - "**.md"
+    paths:
+      - "tidal-listener/**"
   pull_request:
-    paths-ignore:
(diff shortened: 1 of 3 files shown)


ndr-brt (commented):
> Therefore it does not make sense to me to run the full CI (which works towards a packaged release, I think) on every change in Tidal or tidal-link. If I'm wrong, feel free to just close this :)

`tidal-listener` depends on `Tidal`, so technically it makes sense to run the listener CI jobs if something in the root folder change.

the fact that the listener is currently not working is a different topic :)
it deserves some true love

sss-create:
@ndr-brt From this point of view, it certainly does make sense to keep the build-listener active.

PR #1115, opened 2025-01-23
Bump upper bounds

yaxu:
With https://github.com/dktr0/Haskellish/pull/4 , hopefully this will fix builds on latest ghc


sss-create:
Nice! It works!

PR #1120, opened 2025-01-24
Fix upload artifact

yaxu:
Not yet fixed..

### DIFF
--- a/.github/workflows/listener-build-linux.yml
+++ b/.github/workflows/listener-build-linux.yml
@@ -42,7 +42,7 @@ jobs:
     - name: cabal update and build
       run: |
         cabal update
-        cabal build --enable-tests tidal-listener
+        cabal build --enable-tests --allow-newer tidal-listener
         cabal build
 
     - name: move GHC libs and configs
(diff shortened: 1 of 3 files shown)


yaxu:
I upgraded some actions but still it's not working, failing here with a path error:
  https://github.com/tidalcycles/Tidal/actions/runs/12951179422/job/36125648358?pr=1120

I guess something moved with the new action versions. @polymorphicengine I don't suppose you have any time to look at it?

Otherwise we can take it out of the actions for now.

sss-create:
I tackled this a few day ago and traced the .ghcup folder on my machine (linux). The first path error is quite easy to resolve but ‚moving installed package files‘ was unsolvable in my attempt.

polymorphicengine:
i think we should remove the builds, i'm pretty sure even if the build would succeed, the binaries would be broken.. it is still the same linking issue that was introduced by tidal-link and seems to be incompatible with how hint works, here is a reference:
https://github.com/haskell-hint/hint/issues/156

i have to say that at this point i lost the hope of a distributable release of tidal..

yaxu:
> i have to say that at this point i lost the hope of a distributable release of tidal..

A shame given all the work you put into it, but it does seem to be a never-ending nightmare.. Having the haskell interpreter working in the web browser is on the ghc roadmap, at least.

yaxu:
Ok I've disabled the workflows in github. Will merge this anyway, if only to test we get a green build finally..

PR #1126, opened 2025-01-27
fix `beat` function for "step sequencer" style notation (error on last fractional notes)

geikha:
this was just pushed on #1109 

but there was a loose edge case, that is, fractional notes on the upper bound of the divisor. `compress` can't work on ends bigger than `1`, so I added a `min` operator which would squeeze them to fit

see:

```hs
__beat :: (Pattern (Pattern a) -> Pattern a) -> Time -> Time -> Pattern a -> Pattern a
__beat join t d p = join $ (compress (s,e) . pure) <$> p
                      where s = t' / d
                            e  = min 1 $ (t'+1) / d
                            t' = t `mod'` d
```

this error is also present in strudel


yaxu:
Sorry I seem to have really messed this up with the formatting etc. Do you mind redoing the PR please?

geikha:
sure!

PR #1127, opened 2025-01-27
add ormolu job to check formatting

yaxu:


sss-create:
awesome! while i like my leading commas, overall formatting rules are a great improvement  :)

yaxu:
Yes I wouldn't have made the same choices, bit good to not have to make a choice :)
I left a bit of a mess in commit history, tried to tidy it up a bit..

Issue #980, opened 2022-12-21
Mininotation should support top-level `,` and `|`


matthewkaney:
**Describe the function/feature you'd like**

Currently Strudel supports mininotation strings such as `"0, 1"` or `"0 | 1"` while Tidal requires those operators to be wrapped in brackets. The Tidal parser should support this.

More of a stretch: it might be nice to support the combined use of these operators. I'd propose that stacking have a lower precedence than random choice, that is: `"0 | 1, 2 | 3" == "[0 | 1], [2 | 3]"`

Potentially a part of #952

yaxu:
Ref https://github.com/tidalcycles/strudel/issues/284

polymorphicengine:
i can implement this and add it to the open PR for the parser when i have time

sss-create:
@polymorphicengine, how is the status on this? I'd try myself  :)

polymorphicengine:
hey there! currently working on other things, so go for it! :)

i can review the PR then if you'd like

yaxu:
Just noting that at some point it would be nice to collect mininotation tests in a common format that can be tested in both tidal and strudel

sss-create:
It looks like Strudel does not know when the randomization is over. Intuitively, I would expect that `G C` are no longer part of the randomization process:

```
$: note("C|D|G C F")
  .fast(4)
  .sound("sawtooth").cpm(30).gain(.4)
```

but this is `[C | D | G C F]` and not `[C | D | G] C F `

I think to achieve the more intuitive result here must be signifcant whitespace after ```|```. Alternatively should ```|```  be limited to just two numbers/notes/whatever on top level?
Any hints or things I overlook?

PR #1128, opened 2025-01-28
top level '|'

sss-create:
As requested in #980. So far implemented '|' only. I want to make sure I am on the right track. @polymorphicengine, if you would be so kind to take a look :)

I didn't come up with a good way to test, other than (not) throwing exceptions. 

It works and sounds good in the Tidal repl.


yaxu:
Nice! Could you test by comparing the output of e.g. `fast 16 $ "[a a|b b b|c c c c]"` with `fast 16 $ "a a |b b b|c c c c"`? I think they should be the same.

sss-create:
Added the test and it suceeded. Would be so nice if this is really stable. :) Having to put '|' in brackets was one of the first things I thought feel weird, when I started using Tidal. Good to be able to help!

yaxu:
Super thanks a lot!

PR #1133, opened 2025-01-30
Clean warnings

yaxu:


yaxu:
@sss-create Just checking you're not already working on this?

yaxu:
That was easier than I thought!

sss-create:
nice! I started working on this in my fork but did not get too far yet. Did you replace the head/tail functions yet? I thought we could maybe use Debug.Trace to message the user in case of empty list. What is your take on this?

yaxu:
Hm I'm not seeing any warnings about head/tail functions.. How are you getting them?

yaxu:
Ah I see them with the latest rather than the recommended ghc, along with some other new warnings. I don't have any strong opinions about head/tail really.

sss-create:
head/tail warnings are annoying imo. However a strategy to get rid of them is needed. On one hand, the exception head and tail may throw are good to inform the user about a mistake when e.g. passing an empty list. OTOH, it's a scary error message. Same goes for '!!'.

So we'd have to decide what we want to replace head/tail with:
Data.uncons, pattern matching, listToMaybe, NonEmpty or whatever.

yaxu:
I think tail can probably be replaced with `drop 1` in most/all cases. Maybe we don't need to warn the user. There might be some cases where it could be useful but it might result in spamming the console if it's triggered each time a pattern is queried.

sss-create:
I'm undecided but I'd be in favor of removing head/tail from the code base since Tidal's install procedure exposes the user to GHC's warnings. No warning/exception when evaluating a pattern could be irritating though.
_do i use the apostrophs correctly?_ :eyes:

PR #1140, opened 2025-02-06
change link version and bounds to conform to the PVP

polymorphicengine:


yaxu:
I thought worth matching mtl upper bounds in tidal.. I think feel free to merge if the tests pass.

PR #1146, opened 2025-02-10
WIP: top level stack pattern

### LINKED ISSUE #980: Mininotation should support top-level `,` and `|`

sss-create:
so far, the first pattern in the top-level stack has to be a single value
eg: `1 2, 3 4` results in `1 [2, 3 4]`. Where in Strudel it would become `[1 2, 3 4]`
also: no elongate or repeat working in first to be stacked (single) pattern

maybe @polymorphicengine could give a comment on the overall approach.

would resolve #980.

### DIFF
--- a/src/Sound/Tidal/ParseBP.hs
+++ b/src/Sound/Tidal/ParseBP.hs
@@ -403,9 +403,9 @@ pSequence f = do
         spaces
         pEnumeration f a
           <|> pChoice f a
+          <|> pStack f a
           <|> pElongate a
           <|> pRepeat a
-          <|> pStack f a
           <|> return a
         <|> pFoot
   pRand $ resolve_feet s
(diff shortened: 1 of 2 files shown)


sss-create:
I assume that modifying the left-hand side of a `,` (and, I believe, the same applies to top-level `|`) would require a more significant intervention into the and parsing mechanisms? Potentially involving the decoupling of them from the bracket, curly brackets, and parentheses parsing?

yaxu:
Would it work to just generalise `pPolyIn` by removing `brackets $` from it, and use that as the entrypoint for the parser?

sss-create:
@yaxu, that could be an idea! I will try this tomorrow, time to go to bed soon :sleeping:

sss-create:
@yaxu, your suggestion did work! Will merge this now. One more thing to do is to support the combined use of `|` and `,` as described in #980. I'll look into that. Maybe you already can think of a way to support this, again?

PR #1153, opened 2025-02-13
Remove deprecated functions from code base

sss-create:


yaxu:
I'm getting a bit nervous about calling this 2.0 now. You're right that it's an opportunity to make breaking changes like removing deprecated functions, but it feels like with an opportunity like this we should make more far-reaching breaking changes, with a proper ground-up appraisal of everything.

Otherwise it feels like inconveniencing people with incremental changes without enough payback in terms of a tangibly cleaner language overall.

So maybe we should go back to calling it 1.10 with minimum breaking changes, but collect more ambitious plans for a 2.0?

sss-create:
sure, what‘s on your mind for 2.0 then? Yeah, inconveniencing people once would be the idea. 

Perhaps creating a discussion for this?

yaxu:
Heh well having some quite wild thoughts at the moment https://doc.patternclub.org/87ZlbW8NSbWUrkzRRz0SjA?both

We could continue on the 1.10 discussion but expand it to 1.10 vs 2.0 ..

PR #1155, opened 2025-02-16
Extract Core Functionalities into tidal-core Package

eilseq:
This PR reorganizes the Tidal repository by extracting core functionalities into a separate `tidal-core` package to simplifies dependency management, and prepares the codebase for better integration with external tooling that does not depend from stream related functionalities (e.g., Tidal FFI bindings).  

#### **Before (`tidal-core` not separated):**  
- `src/Sound/Tidal/` contained all core logic, including:  
  - `Safe/` (boot and context management)  
  - `Stream/` (event handling and processing)  
  - Various modules like `Pattern.hs`, `Core.hs`, `Params.hs`, `Scales.hs`, `ParseBP.hs`, `Control.hs`, `Utils.hs`, etc.  

#### **After (`tidal-core` extracted):**  
- `src/Sound/Tidal/` now contains only high-level components focusedd on scheduling tasks and networking, while:  
  - `tidal-core/` is introduced as a dedicated package for parsing and construct data structures, designed as follows:  
    - `src/Sound/Tidal/Pattern/`  
      - `Types.hs`, `Pattern.hs`, `Scales.hs`, `ParseBP.hs`, `Stepwise.hs`, `Show.hs`, etc.  
    - `Core.hs`, `Params.hs`, `Control.hs`, `Utils.hs`  
  - A separate `tidal-core.cabal` defining its independent dependencies.  
- This change decouples fundamental Tidal pattern processing from streaming and external integrations (e.g., Ableton Link).  

### **Why This Change?**  
- **Modularization:** Separating core logic enables better reusability and maintainability.  
- **Dependency Isolation:** `tidal-core` can be built/tested independently, avoiding unnecessary dependencies.  
- **Better Cross-Language Support:** Facilitates integration with Rust FFI (`tidal-parse-ffi`) without requiring the entire Tidal stack.  
- **Reduced Build Complexity:** Removing Ableton Link and other non-essential dependencies from the core simplifies build configurations for external targets based on core functionalities.


yaxu:
This sounds good in principle. I guess install commands and use for conventional tidal users won't be impacted?

Strudel also has a standalone `core` package for the pattern engine, with no dependencies apart from a library for rational numbers. It also has the mininotation parser as its own package.

I think @polymorphicengine and @matthewkaney have been involved in splitting things out in tidal recently and might be better to comment on the specifics, if they have time..

eilseq:
Yes, **no breaking changes**:
_tidal-core_ in this context would be a base dependency in the monorepo, so that _tidal_ consumption would stay the same

eilseq:
Thank you for taking the time for reviewing this @yaxu!
Took me some attempts, but now the building pipeline seems to be working.

I still see an issue with the job ormolu, but I could not figure out the reason based on the error prompt.

yaxu:
Good stuff! The main question is I think, how to handle upgrades and interdependencies.

Currently `tidal-core` is a dependency of `tidal` without bounds. That suggests `tidal-core` will be compatible with all versions of `tidal`. According to the [PVP](https://pvp.haskell.org/) (haskell package versioning policy) `tidal-core` should have bounds like `>=1.10 && <1.11`, to indicate it's compatible with future tidal-core versions without breaking changes.

However I guess it will be simplest for end-user live coders if we keep version numbers consistent across `tidal` and `tidal-core`, so if you up/downgrade `tidal` then you will also get the `tidal-core` with the same version number. We might release a new `tidal` without a new `tidal-core`, but when we release a new `tidal-core` we should always release a new `tidal` with the same version number.

tl;dr think the bounds `==1.10` should be added to the `tidal-core` dependency of `tidal`

yaxu:
Ok I made that change and merged, thanks!
We can upload tidal-core to hackage first, ahead of a new tidal release.

yaxu:
PR opened for inclusion in stackage:
https://github.com/commercialhaskell/stackage/pull/7689

PR #1157, opened 2025-02-19
Migrate from Microspec to Hspec

sss-create:
as discussed in #1130, Microspec was last updated in 2018. Therefore, I substituted it with Hspec, which is very well maintained.

@yaxu, please review and merge if you find everything ok :)


yaxu:
Great stuff!

Issue #813, opened 2021-03-30
Check for dangerous and unsafe functions


yaxu:
Useful list and hlint config here: 
https://github.com/NorfairKing/haskell-dangerous-functions

jwaldmann:
I am looking into this (as a way of refreshing my knowledge of Tidal's code...) NB: https://github.com/NorfairKing/haskell-dangerous-functions/issues/18

jwaldmann:
With this "dangerous function" hlint.yaml, I get this report on Tidal's current main branch: https://www.imn.htwk-leipzig.de/~waldmann/etc/Tidal/

This reports all  warnings (including those from hlint's default settings) I guess that "dangerous functions" are listed under "avoid restricted function".

NB: It's a bit strange (for me) to see "return" there, but apparently it should be "pure". That's a matter of style only? No, this page https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/monad-of-no-return says it's "error prone" (default declarations of Applicative.pure and Monad.return refer to each other => non-termination).

PR #1162, opened 2025-02-21
link and clock setter functions

sss-create:
the setters allow to define clock latencies via, as discussed in https://github.com/tidalcycles/tidal-doc/issues/289. I swapped the arguments for `mkTidalWith` to align with previous tidal versions.  
`tidalInst <- mkTidalWith mkOscMap (setFrameTimespan (1/20) $ setProcessAhead (3/10) defaultConfig) `

also `enableLink` and `disableLink` were added and commented in the `BootTidal.hs`


sss-create:
oh yeah, while I was at it i changed some `return`'s to `pure` in the spirit of #813

PR #1193, opened 2025-05-01
Corrects the scaleWithList function 

aherrou:
It now behaves in a more intuitive way, with the cycles underlying the functions list synchronizing with the pattern's cycles instead of duplicating them. Implementation inspired by that of `cat`.

(sorry for the duplicate pull request, I forgot to merge the main branch before adding my contributions)


yaxu:
Thanks!

PR #1204, opened 2025-06-06
import IsString from a portable module

claudeha:
`GHC.Exts` is GHC-only, `Data.String` works with other compilers too (like MicroHs)


yaxu:
Thanks!

PR #1206, opened 2025-06-07
Revert "Corrects the scaleWithList function "

yaxu:
Reverts tidalcycles/Tidal#1193

There's something funny going on here.. More things missing from Sound.Tidal.Pattern, but when I import everything with `import Sound.Tidal.Pattern`, `scaleWithList` doesn't typecheck. Could you take a look please @aherrou  ?

### DIFF
--- a/tidal-core/src/Sound/Tidal/Scales.hs
+++ b/tidal-core/src/Sound/Tidal/Scales.hs
@@ -20,7 +20,7 @@ module Sound.Tidal.Scales (scale, scaleList, scaleTable, getScale, scaleWith, sc
 
 import Data.Maybe (fromMaybe)
 import Sound.Tidal.Core (slowcat)
-import Sound.Tidal.Pattern (Pattern, arc, (<*))
+import Sound.Tidal.Pattern (Pattern, (<*))
 import Sound.Tidal.Utils ((!!!))
 import Prelude hiding ((*>), (<*))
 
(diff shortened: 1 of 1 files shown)


yaxu:
Looking at the underlying issue with the original version of it:
```hs
scaleWithList sp fs p = slowcat $ map (\f -> scaleWith sp f p) fs
```
`chunk` works in a same way, repeating the source pattern for each variant, and it can be quite nice. In strudel we made `fastchunk` which doesn't repeat cycles from the source pattern (note to self - backport that to tidal). I haven't used it myself, but are there cases when someone might want the previous, 'slow' behaviour of `scaleWithList`?

In any case I feel there should be a shorter way of expressing the desired behaviour without getting into the weeds of manipulating events directly, will have a think..

yaxu:
I haven't tested it (not sure how it's actually used!) but how about this:
```hs
scaleWithList sp fs p = innerJoin $ (\f -> scaleWith sp f p) <$> slowcat fs
```
Then if the old behaviour is useful to someone we could preserve it as `scaleWithListSlow`:
```hs
scaleWithListSlow :: (Eq a, Fractional a) => Pattern String -> ([[a] -> [a]]) -> Pattern Int -> Pattern a
scaleWithListSlow sp fs p = slowcat $ map (\f -> scaleWith sp f p) fs
```
Does that make sense @aherrou ?

